PR #19, opened 2020-07-11
Improved tool selection

Slord6:
I've updated bestHarvestTool to better select tools. It should, on average, choose the best available harvest tool now. Also, if there is no harvest tool associated with a block, eg `grass`, `leaves`, it will select a tool based on the block material

Edit: Worth noting that still does not cover enchanted tools

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,13 +39,13 @@ function inject (bot) {
   let thinking = false
   let lastNodeTime = performance.now()
 
-  function resetPath () {
+  function resetPath (clearStates = true) {
     path = []
     if (digging) bot.stopDigging()
     digging = false
     placing = false
     pathUpdated = false
-    bot.clearControlStates()
+    if (clearStates) { bot.clearControlStates() }
   }
 
   bot.pathfinder.setGoal = function (goal, dynamic = false) {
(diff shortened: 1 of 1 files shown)


Karang:
Thanks for this! It is indeed something that needs improvement.
We now have an accurate digTime() function in prismarine-block that could be used to implement the best tool selection.
Here is an example: https://github.com/Karang/mineflayer-pathfinder/blob/master/lib/movements.js#L108
This support the real speed modifier of the item as well as enchantments and player abilities and is guaranteed to be updated for future versions. Could you try to use it ?

Also, make sure you rebased on master, you seems to be missing some of the latest changes (look at the diff of the PR).

Slord6:
Didn't see digTime, I'll have another look and refactor, that should make it more robust.

Not sure how I ended up with an old version, only forked the repo yesterday! Thanks for the heads up.

Slord6:
I've made a complete mess of that feature branch somehow. Will close this PR and reimplement as per your above comment!

Karang:
> I've made a complete mess of that feature branch somehow. Will close this PR and reimplement as per your above comment!

Thanks!

PR #21, opened 2020-07-11
Tool selection improvements

Slord6:
As per discussion on PR #19: Improved tool selection using `digTime`.

I'm open to suggestions on improving the new `itemIsTool` check, feels like there should be a better way to check than string comparisons, but I think it works as a first pass anyhow.

Tested with the `example/test.js` bot on `1.16.1` and seems to work

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,47 +6,42 @@ const Move = require('./lib/move')
 const Vec3 = require('vec3').Vec3
 
 const { PlayerState } = require('prismarine-physics')
+const nbt = require('prismarine-nbt')
 
 const THINK_TIMEOUT = 40 // ms
 
 function inject (bot) {
   bot.pathfinder = {}
 
-  function bestToolOfTypeInInventory (bot, toolname, materials) {
-    const tools = materials.map(x => x + '_' + toolname)
-    for (let i = tools.length - 1; i >= 0; i--) {
-      const tool = tools[i]
-      const matches = bot.inventory.items().filter(item => item.name === tool)
-      if (matches.length > 0) return matches[0]
+  function itemIsTool(item) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -6,22 +6,41 @@ const Move = require('./lib/move')
...
+  }
+
   bot.pathfinder.bestHarvestTool = function (block) {
-    const items = bot.inventory.items()
-    for (const i in block.harvestTools) {
-      const id = parseInt(i, 10)
-      for (const j in items) {
-        const item = items[j]
-        if (item.type === id) return item
+    if (block.name === 'air') return null
```
Karang, line 27: You could use `block.type === 0` id 0 is always air for all mc versions.
But is it really needed ? Does it happen often that bestHarvestTool gets called with an air block as a parameter ?

index.js
```diff
@@ -6,22 +6,41 @@ const Move = require('./lib/move')
...
+
+    const availableTools = bot.inventory.items().filter(itemIsTool)
```
Karang, line 29: digTime should work with any item and tools should have the fastest dig time anyway. Is it really needed to prefilter the items ?

index.js
```diff
@@ -6,22 +6,41 @@ const Move = require('./lib/move')
...
+    const effects = bot.entity.effects
+
+    let fastest = Number.MAX_VALUE
+    let bestTool = null
+    availableTools.forEach(tool => {
```
Karang, line 34: please use:
```js
for (const tool of availableTools) {
  //
}
```

index.js
Slord6: Think this was left over from when I was using the functionality elsewhere, I'll remove it

Karang:
Merged and released. Thanks

PR #22, opened 2020-07-12
Modular Movements Config Proposal

TheDudeFromCI:
This is the first revision of a basic configuration file for the new modular movement system. This should add all movement functionality for walking, mining, and bridging horizontally. Does not yet support jumping or falling.


Karang:
Thanks.
One thing we have in the original design is the concept of "optional actions" maybe translating this concept here would simplify things. We also need to specify conditions for if an action cannot be made. For instance: 
```
{
  "type": "mine",
  "actionCondition": "not air", // do this action only if true
  "condition": "canBreak", // if false, the move is cancelled
  "delta": [0, 0, -1]
}
```
(Maybe find better names for the 2 conditions)

The action condition could make all "mine forward xxxxx" one move in the specification.

For the forward direction it is ok to pick a convention, but for radial symmetry it doesn't matter. Here is the code I had to generate all directions (works with diagonals too): 
```js
function radialSymetry (delta) {
  return [[delta[0], delta[1], delta[2]], [-delta[2], delta[1], delta[0]], [-delta[0], delta[1], -delta[2]], [delta[2], delta[1], -delta[0]]]
}
```

Karang:
Its starting to take shape :) What do you think of having higher level conditions like 'safeToBreak' or 'safeToWalkIn' that can do additional checks than just bounding box (like checking if one of the neighbor is a liquid that have a risk to spread) ? Alternatively, we would need to add simple conditions like "nextToLiquid". Or is it something that should be implicit ?

You can also start to implement a basic interpreter for the format (something very basic, no optimizations yet) this would allow to verify if the format is expressive enough and start testing.

An other concept we could include is the dependency between moves. For instance, if we generate a jump over a 1 block gap, we don't need to generate a move bridging over the gap, because we know it will cost more. This can be done by having an explicit dependency between moves:
```
{
      "name": "forward",
      "dontDoIf": "jump forward over gap"
      "symmetry": "radial",
      "actions": [ ...]
}
```

Karang:
Yes, gravity block should be considered as well. You are right about the path being recomputed, but liquid flow is generally something we want to avoid, so we should be able to (it is currently configurable with default to true).

The movements dependency is purely to optimise the number of node generated (this is why I said its optional). Because that something that cannot be automatically detected and optimised at the generator level.

Karang:
Looks interesting. I have a few comments but i want to test it first. I'll review asap

Karang:
Could you adapt the api to be compatible with the current one, without modifying astar or index ? This is important to be able to swap the 2 when checking the performances. We should focus on 1 feature at a time to make sure it improves the code.

Karang:
That would do it :) thanks

lib/mod_movements/handler.js
```diff
@@ -0,0 +1,170 @@
...
+
+      if (!this._meetsRequirements(node, world, requirements)) continue
+
+      const block = world.virtualBlockAt(node, node.position + delta)
+      const a = new Action(type, block.position)
+
+      if (type === 'mine') {
+        if (!this._canMine(block)) continue
+
+        a.tool = 'hand' // TODO Determine best tool
```
Karang, line 98: This is better left as a runtime decision. It doesn't impact which path is better, you are only interested to know whether or not you can break the block, you don't need to store the tool you need. Memory is precious and checking the inventory for best tool is fast enough to be redone when walking the path.

lib/mod_movements/handler.js
```diff
@@ -0,0 +1,170 @@
...
+      }
+
+      if (type === 'build') {
+        if (!this._canPlace(block)) continue
+
+        a.scaffolding = 'dirt' // TODO Determine correct scaffolding
```
Karang, line 104: Same comment as for the digging, you don't need to store that information in the action. You only want to know if you have scaffolding blocks remaining, you don't care which one.

lib/mod_movements/handler.js
```diff
@@ -0,0 +1,170 @@
...
+    else return null
+  }
+
+  /**
+   * Generates the hash string for a movement object.
+   *
+   * @param {*} hash - The hash string.
+   * @param {*} position - The position of the movement.
+   */
+  _generateHash (hash, position) {
```
Karang, line 124: I don't think the hash should be configurable by the user. It is is important for the algorithm that the hash is a position so multiple node pointing to the same block would overwrite each other. Using replace will slow the whole algorithm compare to the current simple concatenation.

lib/mod_movements/virtualWorld.js
```diff
@@ -0,0 +1,76 @@
...
+    this.world = world
+  }
+
+  /**
+   * Gets what block would be at this location at this node along the path.
+   *
+   * @param {Movement} node - The current movement node to check.
+   * @param {*} position - The position within the world.
+   */
+  virtualBlockAt (node, position) {
```
Karang, line 27: I get what you want to do here, but I don't think it is worth it. The current algorithm already assume that the position the block is standing in is valid and most of the time it will not revisit previous nodes. In addition, paths are recomputed each time the environment changes (including when the bot modifies it). The block access needs to be kept fast, so this is not something you want to change unless there is a very good reason for it.

Can you provide an example where taking past actions into account gives a better quality path than what is currently implemented ?

lib/mod_movements/handler.js
```diff
@@ -0,0 +1,170 @@
...
+
+      m.actions.push(a)
+
+      if (a.effectsBotPosition()) m.position = action.position
+    }
+
+    m.hash = this._generateHash(movement.hash, m.position)
+
+    if (m.actions.length > 0) return m
+    else return null
```
Karang, line 115: You do not set the cost of the movement.

lib/mod_movements/virtualWorld.js
TheDudeFromCI: Yes, a common example of this, including addressing the configurable hashing implementation, would be for world-editing based tasks. Namely, where the goal of the task is not to get to a certain position but to edit a certain number of blocks. Storing what blocks where modified and how is essential.

Another example is used when handling combined goals where two objectives must be met. This may include breaking a block to access one of the waypoints and replacing it to bridge across to another.

One thing that concerns me is the bot recalculating the path when modifying it's own blocks. This leads to the bot "forgetting" information such as waypoints like mentioned above.

At current, the bot is only capable of moving from one destination to another with little room for configuring extra actions to be preformed along the way.

lib/mod_movements/handler.js
TheDudeFromCI: While I agree that should be the default option, I think considering this information to be considered at solve time is also useful if the developer decides to turn that on. It can be much easier to travel a few extra blocks and tunnel through a dirt wall than try and mine through an Obsidian wall with your hand.

Karang: It is already considered as part of the cost of the movement. But you don't need to store the tool name.

lib/mod_movements/virtualWorld.js
Karang: For both cases, it is more efficient to address the issue at a higher level. When I was asking for an example, I meant a real (implemented) example on which you run both versions and compare the output.

In other words, the waypoint and building use cases are both very interesting, but they are out of scope of the modular movement system (and there is more than one way to implement them). IMO there is no need to complicate and slow things now for features we don't need yet.

PR #23, opened 2020-07-12
Interact Goal

TheDudeFromCI:
Adds a new goal which puts the bot into a position where it can interact with a target block. This is similar to `GoalGetToBlock`, however, does not require the bot to be near the target block. This allows the bot to interact with blocks from a distance or mine specifc blocks which obstruct it's view to get to the block.

Some examples of valid positions are: (`=` is stone, `B` is bot, `X` is target)
```
==== =
B   X=
B ====
======
```

```
B
B    X
=    =
```

```
X

B
B
=
```

```
B
B X
===
```

### DIFF
--- a/lib/goals.js
+++ b/lib/goals.js
@@ -251,6 +251,79 @@ class GoalFollow extends Goal {
   }
 }
 
+// Moves the bot to a position where it is able to interact with the target
+// block, even from a slight distance. Used to create a more natural approach
+// to interacting and mining without needing to be so close to the target.
+class GoalInteract extends MoveTo {
+  constructor (bot, x, y, z) {
+    super(x, y, z)
+    this.bot = bot
+  }
+
+  isEnd(node)
+  {
+      return this.rayTraceBlock(node);
+  }
+
+  rayTraceBlock(node, maxSteps = 256, vectorLength = 5 / 16) {
+    const height = 1.6; // TODO Check if bot is sneaking at this node
+
+    const start = new Vec3(node.x + 0.5, node.y + height, node.z + 0.5);
+    const end = new Vec3(this.x + 0.5, this.y + 0.5, this.z + 0.5);
+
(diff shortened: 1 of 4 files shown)


lib/goals.js
```diff
@@ -251,6 +251,79 @@ class GoalFollow extends Goal {
...
+    super(x, y, z)
+    this.bot = bot
+  }
+
+  isEnd(node)
+  {
+      return this.rayTraceBlock(node);
+  }
+
+  rayTraceBlock(node, maxSteps = 256, vectorLength = 5 / 16) {
```
Karang, line 268: You can use the RaycastIterator I added to mineflayer: https://github.com/PrismarineJS/mineflayer/blob/master/lib/iterators.js#L105
It computes the step so that each step advance exactly to the next block (so you don't need ridiculously small steps to not ghost through blocks), resulting in only 1 check per block on the line of sight.

TheDudeFromCI: I can't use existing raycasters as the goal relies on what the state of the world *will* be at that point in time. The blocks to be broken haven't been mined yet, so the world raycaster will deny it even though it should pass.

Karang: The iterator is not tied to a world, it only iterates on block positions. So you can use it in the context you want.

TheDudeFromCI: Oh! How did I miss this? Wow, that's significantly more helpful, lol. I'll definitely be using this, then.

At first, I thought you were talking about this one: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/ray_trace.js

Karang: yeah,  https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/ray_trace.js should be replaced as well

Karang:
@TheDudeFromCI what's the status for this ? Should we close it for now or do you still plan to work on it ?

PR #35, opened 2020-07-28
add simple params to specify host like in mineflayer examples

rom1504:


Karang:
Could you add default values for host and port so it can connect to localhost if no params are given ?

rom1504:
done

PR #43, opened 2020-08-29
Update index.d.ts

RayhanHamada:
Hi there.
I'm using typescript, and it's seems intellisense won't suggesting pathfinder function when i import this package, so i add export for pathfinder and this seems fix the problem. and i'm also fix other types like the one inside Pathfinder class so it won't be inferred to any.


Karang:
@TheDudeFromCI can you review this ?

PR #58, opened 2020-11-01
refactor: refactor module

NotSugden:
First of all, I realise there are some major code style changes here, if the maintainers want i can revert some of those style changes

This PR does a major refactor to this module, here is a list of changes

**anyone who can help test this please do!, I have encountered issues with the bot jumping and placing a block underneath it, im not sure if this is an existing issue or something as a result of a change i've made**

- bumps the package version to `2.0.0` as there are major (breaking) changes in this PR
- adds sprinting funcionality
- adds (correct) typings
- refactors the injection function to:
  - renamed to `inject`
  - uses a class instead of how it was previously done
- code style changes:
  - ES6 arrow function syntax
  - uses JSDoc for comments on specific classes where applicable
  - refactors some functions to use built-in array methods like `Array.reduce` and `Array.some`


rom1504:
> bumps the package version to 2.0.0 as there are major (breaking) changes in this PR

is this required ?

NotSugden:
> > bumps the package version to 2.0.0 as there are major (breaking) changes in this PR
> 
> is this required ?

well i suppose i could change the behaviour so its not breaking, but ultimately it feels better to just bump the major version instead, its up to you though

- alter this PR so no existing behaviour is broken
- keep the breaking changes and bump the major version

rom1504:
What are the breaking changes and what's the benefit of them being breaking?
Introducing a major version has a big cost as it would require updates in all dependent packages and repos. So it makes sense to do that only if the breaking changes are worth it.

NotSugden:
> What are the breaking changes and what's the benefit of them being breaking?
> Introducing a major version has a big cost as it would require updates in all dependent packages and repos. So it makes sense to do that only if the breaking changes are worth it.

well i wasn't sure about every single breaking change as i didnt note them, but i'll compare everything and see

rom1504:
I'd advise separating the new features and the refactoring changing nothing in 2 separate PRs, this looks hard to review

NotSugden:
@rom1504 

- bot.pathfinder is now a class, and not an object with various functions attached
- the inject function is now exported as `inject` and not `pathfinder`
- `bot` is now read-only and non-enumerable on the `Movements` class
- removed `pathfinder.isMining()` in favour of `pathfinder.mining`
- removed `pathfinder.isBuilding()` in favour of `pathfinder.building`
- removed `pathfinder.isThinking()` in favour of `pathfinder.thinking`
- removed the `done` callback parameter from `pathfinder.getPathTo()` as it was unnesercary
- `Movements#getBlock()` now returns its own object rather than directly modifying the `Block` recieved from `minecraft-data`
  - this change has lead to various parameter changes where a `Block` was previously accepted

NotSugden:
> I'd advise separating the new features and the refactoring changing nothing in 2 separate PRs, this looks hard to review

i'll look into doing this, although i don't think there are many intended new features in here, off the top of my head i added sprinting support (although that may warrant its own PR, i don't know if there are any side-effects from sprinting that will need to be addressed seperately)

Karang:
> @rom1504
> 
>     * bot.pathfinder is now a class, and not an object with various functions attached
> 
>     * the inject function is now exported as `inject` and not `pathfinder`
> 
>     * `bot` is now read-only and non-enumerable on the `Movements` class
> 
>     * removed `pathfinder.isMining()` in favour of `pathfinder.mining`
> 
>     * removed `pathfinder.isBuilding()` in favour of `pathfinder.building`
> 
>     * removed `pathfinder.isThinking()` in favour of `pathfinder.thinking`
> 
>     * removed the `done` callback parameter from `pathfinder.getPathTo()` as it was unnesercary
> 
>     * `Movements#getBlock()` now returns its own object rather than directly modifying the `Block` recieved from `minecraft-data`
>       
>       * this change has lead to various parameter changes where a `Block` was previously accepted

I don't get why you need to break the api?
I'm going to comment directly on the list of feature :
* what's the advantage of having pathfinder a class instead of a singleton object ?
* it was more explicit to be able to do `bot.loadPlugin(pathfinder)` instead of `bot.loadPlugin(inject)`, please revert
* what's the advantage of this ? Movement is an internal class anyway.
* this is an api break, what is the justification for it ?
* same
* same
* the done callback was intended to be used when implementing the async computation (if it takes more than a tick to compute)
* this was done like this for performance reasons: creating too many objects is expensive, recycling is better for this sensitive function

Keep it mind that this code is very performance critical, so when introducting that much changes, it will have to be benchmarked again and compared with the current version.

NotSugden:
@Karang 

- a class is alot easier to read and document than just a single object
- fair point, will revert
- i thought the `Movements` class is apart of the public API? the user instantiates it and the user changes properties on it to configure it ?
- well, seeing as the major version was going to be bumped due to other changes, i opted to remove this in favour of a getter as the properties exist on the class itself now there isn't a need for them to be a function, there isn't really a need for them to be aliased at all really
  - `pathfinder.isMining()` would be the same as just `pathfinder.digging`, however `pathfinder.isMining()` has been removed and now `pathfinder.mining` is just an alias for `pathfinder.digging`
- same
- same
- its not being used at the moment though ? shouldn't that have all gone into one PR? if im understanding what you're saying this was a WIP?
- but properties are being added to the `Block` object that shouldn't be there, couldn't this break behaviour if someone did say `Object.keys(minecraftData.blocksById[...])` (this is just an example), either way it seems a bit silly to me to do it that way, but if you do insist i can revert that

lib/goto.js
```diff
@@ -13,23 +13,21 @@ function error (name, message) {
...
-    cleanup(error('GoalChanged', 'The goal was changed before it could be completed!'))
+    cleanup(
+      error('GoalChanged', 'The goal was changed before it could be completed!')
+    )
   }
 
-  function cleanup (err) {
-    bot.removeListener('goal_reached', goalReached)
+  const cleanup = err => {
+    bot.removeListener('goal_reached', cleanup)
```
TheDudeFromCI, line 30: The goal reached event does pass a param to the callback, so using cleanup as the listener will wrongly throw an "error."

TheDudeFromCI (changes requested):

lib/goto.js
NotSugden: oh, that is indeed correct, my bad

Karang:
> > I'd advise separating the new features and the refactoring changing nothing in 2 separate PRs, this looks hard to review
> 
> i'll look into doing this

You really should do what Rom suggested. I don't get why you are doing the "change style" part of this PR: as far as I'm concern there are some interesting stuff in this PR, but most of it is just shuffling things around for seemingly no reasons...

NotSugden:
> > > I'd advise separating the new features and the refactoring changing nothing in 2 separate PRs, this looks hard to review
> > 
> > 
> > i'll look into doing this
> 
> You really should do what Rom suggested. I don't get why you are doing the "change style" part of this PR: as far as I'm concerned there are some interesting stuff in this PR, but most of it is just shuffling things around for seemingly no reasons...
> 
> There is also a few changes that I think would slow down the pathfinding, this module was built with speed in mind, if you think your changes can make things faster, you'll have to run some benchmarks.

ok, ill make some branches on my fork and split up this PR a little bit, and then ill run some benchmarks to compare with the current version

examples/bench.js
```diff
@@ -21,15 +20,29 @@ bot.loadPlugin(pathfinder)
 
 const createTime = performance.now()
 bot.once('spawn', () => {
-  console.log('Spawning took ' + (performance.now() - createTime).toFixed(2) + ' ms.')
+  console.log(
+    `Spawning took ${(performance.now() - createTime).toFixed(2)} ms.`
+  )
```
Karang, line 25: No need to make it spawn multiple lines.

examples/bench.js
```diff
@@ -21,15 +20,29 @@ bot.loadPlugin(pathfinder)
...
+      console.log(
+        `I can get there in ${
+          path.length
+        } moves. Computation took ${time.toFixed(2)} ms.`
+      )
+      bot.quit()
+      process.exit()
+    },
+    10000
+  )
```
Karang, line 47: No need to make it spawn multiple lines.

examples/multiple.js
```diff
@@ -1,8 +1,7 @@
 const mineflayer = require('mineflayer')
-const { pathfinder, Movements } = require('mineflayer-pathfinder')
-const { GoalInvert, GoalFollow } = require('mineflayer-pathfinder').goals
+const { goals, pathfinder, Movements } = require('mineflayer-pathfinder')
 
-mineflayer.multiple = (bots, constructor) => {
+const multiple = (bots, constructor) => {
```
Karang, line 4: It was intended like this in prevision of moving this function to mineflayer. Please keep the namespace.

examples/multiple.js
```diff
@@ -19,37 +18,53 @@ for (let i = 0; i < 40; i++) {
...
-    bot.on('goal_reached', (goal) => {
-      console.log('[' + username + '] Here I am !')
-    })
+    bot.on('path_update', ({ path, time }) =>
+      log(
+        `I can get there in ${
+          path.length
+        } moves. Computation took ${time.toFixed(2)} ms.`
+      )
+    )
```
Karang, line 39: No need to make it spawn multiple lines.

examples/multiple.js
```diff
@@ -19,37 +18,53 @@ for (let i = 0; i < 40; i++) {
...
-        bot.pathfinder.setGoal(new GoalFollow(target, 5), true)
+        bot.pathfinder.setGoal(new goals.GoalFollow(target, 5), true)
       } else if (message === 'avoid') {
+        if (!canSee()) return
         bot.pathfinder.setMovements(defaultMove)
-        bot.pathfinder.setGoal(new GoalInvert(new GoalFollow(target, 5)), true)
+        bot.pathfinder.setGoal(
+          new goals.GoalInvert(new goals.GoalFollow(target, 5)),
+          true
+        )
```
Karang, line 67: No need to make it spawn multiple lines.

examples/test.js
```diff
@@ -23,59 +22,67 @@ bot.once('spawn', () => {
...
-  bot.on('path_update', (r) => {
-    const nodesPerTick = (r.visitedNodes * 50 / r.time).toFixed(2)
-    console.log(`I can get there in ${r.path.length} moves. Computation took ${r.time.toFixed(2)} ms (${nodesPerTick} nodes/tick).`)
+  bot.on('path_update', ({ path, time, visitedNodes }) => {
+    const nodesPerTick = ((visitedNodes * 50) / time).toFixed(2)
+    console.log(
+      `I can get there in ${path.length} moves. Computation took ${time.toFixed(
+        2
+      )} ms (${nodesPerTick} nodes/tick).`
+    )
```
Karang, line 31: No need to make it spawn multiple lines.

lib/astar.js
```diff
@@ -32,20 +28,23 @@ function reconstructPath (node) {
...
     this.closedDataSet = new Set()
     this.openHeap = new Heap()
     this.openDataMap = new Map()
 
-    const startNode = new PathNode().set(start, 0, goal.heuristic(start))
+    const startNode = (this.bestNode = new PathNode().set(
+      start,
+      0,
+      goal.heuristic(start)
+    ))
```
Karang, line 45: No need to make it spawn multiple lines.

lib/astar.js
```diff
@@ -96,7 +95,12 @@ class AStar {
         }
         // found a new or better route.
         // update this neighbor with this node as its new parent
-        neighborNode.set(neighborData, gFromThisNode, this.goal.heuristic(neighborData), node)
+        neighborNode.set(
+          neighborData,
+          gFromThisNode,
+          this.goal.heuristic(neighborData),
+          node
+        )
```
Karang, line 103: No need to make it spawn multiple lines.

lib/astar.js
```diff
@@ -110,4 +114,7 @@ class AStar {
   }
 }
 
-module.exports = AStar
+module.exports = {
+  AStar,
+  PathNode
+}
```
Karang, line 120: No need to make it spawn multiple lines.

lib/goals.js
```diff
@@ -1,23 +1,36 @@
...
+/**
+ * Goal base class
+ * @abstract
+ */
 class Goal {
-  // Return the distance between node and the goal
+  /**
+   * Return the distance between node and the goal
+   * @abstract
+   */
```
Karang, line 9: I prefer // over /**/ for comments as it makes it easier to comment codes when developing.
I'm also not a big fan of bulky comments.

lib/goals.js
```diff
@@ -1,23 +1,36 @@
...
-  // Return true if the goal has changed and the current path
-  // should be invalidated and computed again
+  /**
+   * Return true if the goal has changed and the current path
+   * should be invalidated and computed again
+   */
   hasChanged () {
     return false
-  }
+  };
```
Karang, line 28: Extra ;

lib/goals.js
```diff
@@ -1,23 +1,36 @@
...
 
-  // Return true if the node has reach the goal
+  /**
+   * Return true if the node has reach the goal
+   * @abstract
+   */
   isEnd (node) {
     return true
-  }
+  };
```
Karang, line 20: Extra;

lib/heap.js
```diff
@@ -1,7 +1,5 @@
-
 class BinaryHeapOpenSet {
   constructor () {
-    // Initialing the array heap and adding a dummy element at index 0
```
Karang, line 4: Don't remove this comment.

lib/heap.js
```diff
@@ -23,7 +21,10 @@ class BinaryHeapOpenSet {
 
     // Traversing up the parent node until the current node is greater than the parent
     while (current > 1 && this.heap[parent].f > this.heap[current].f) {
-      [this.heap[parent], this.heap[current]] = [this.heap[current], this.heap[parent]]
+      ;[this.heap[parent], this.heap[current]] = [
```
Karang, line 24: Extra ; No need to make it spawn multiple lines.

lib/move.js
```diff
@@ -1,16 +1,27 @@
...
+  constructor (
+    x,
+    y,
+    z,
+    remainingBlocks,
+    cost,
+    toBreak = [],
+    toPlace = [],
+    parkour = false
+  ) {
```
Karang, line 11: No need to make it spawn multiple lines.

lib/move.js
```diff
@@ -1,16 +1,27 @@
...
     this.remainingBlocks = remainingBlocks
     this.cost = cost
     this.toBreak = toBreak
     this.toPlace = toPlace
     this.parkour = parkour
+  }
 
-    this.hash = this.x + ',' + this.y + ',' + this.z
+  get hash () {
+    return `${this.x},${this.y},${this.z}`
```
Karang, line 24: The hash was precomputed for a reason.

lib/movements.js
```diff
@@ -1,94 +1,92 @@
...
-  { x: 1, z: 1 }
-]
+const _minecraftData = require('minecraft-data')
+const { cardinalDirections, diagonalDirections } = require('./constants')
+const { Move } = require('./Move')
 
 class Movements {
-  constructor (bot, mcData) {
-    this.bot = bot
+  constructor (bot, minecraftData = _minecraftData(bot.version)) {
```
Karang, line 8: It is better to encourage people to give mcData rather than creating a new one behind their back. There are plans to have a single instance of mcData available in the bot, so this could be simplified, but not this way.

lib/movements.js
```diff
@@ -1,94 +1,92 @@
...
+    const safeBlock = {
+      safe: block.boundingBox === 'empty' && !this.avoidBlocks.has(block.type),
+      physical: block.boundingBox === 'block',
+      liquid: liquids.has(block.type),
+      height: originalHeight,
+      block
     }
-    return b
+
+    if ('shapes' in block) {
```
Karang, line 83: This test is not necessary.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
       if (this.getBlock(block.position, 0, 0, -1).liquid) return false
       if (this.getBlock(block.position, 0, 0, 1).liquid) return false
     }
-    return block.type && !this.blocksCantBreak.has(block.type)
-    // TODO: break exclusion areas
```
Karang, line 108: Don't remove this comment

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
-    const digTime = block.digTime(tool ? tool.type : null, false, false, false, enchants, effects)
-    return (1 + 3 * digTime / 1000) * this.digCost
+    const digTime = block.digTime(
+      tool ? tool.type : null,
+      false,
+      false,
+      false,
+      enchants,
+      effects
+    )
```
Karang, line 123: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
         if (node.remainingBlocks === 1) return // not enough blocks to place
-        toPlace.push({ x: node.x, y: node.y - 1, z: node.z, dx: dir.x, dy: 0, dz: dir.z })
+        toPlace.push({
+          x: node.x,
+          y: node.y - 1,
+          z: node.z,
+          dx: dir.x,
+          dy: 0,
+          dz: dir.z
+        })
```
Karang, line 151: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
 
-      toPlace.push({ x: node.x + dir.x, y: node.y - 1, z: node.z + dir.z, dx: 0, dy: 1, dz: 0 })
+      toPlace.push({
+        x: node.x + dir.x,
+        y: node.y - 1,
+        z: node.z + dir.z,
+        dx: 0,
+        dy: 1,
+        dz: 0
+      })
```
Karang, line 162: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+      new Move(
+        blockB.position.x,
+        blockB.position.y,
+        blockB.position.z,
+        node.remainingBlocks - toPlace.length,
+        cost,
+        toBreak,
+        toPlace
+      )
+    )
```
Karang, line 190: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
       if (node.remainingBlocks === 0) return // not enough blocks to place
-      toPlace.push({ x: node.x, y: node.y - 1, z: node.z, dx: dir.x, dy: 0, dz: dir.z })
+      toPlace.push({
+        x: node.x,
+        y: node.y - 1,
+        z: node.z,
+        dx: dir.x,
+        dy: 0,
+        dz: dir.z
+      })
```
Karang, line 211: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+      new Move(
+        blockC.position.x,
+        blockC.position.y,
+        blockC.position.z,
+        node.remainingBlocks - toPlace.length,
+        cost,
+        toBreak,
+        toPlace
+      )
+    )
```
Karang, line 232: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+    neighbors.push(
+      new Move(
+        blockC.position.x,
+        blockC.position.y,
+        blockC.position.z,
+        node.remainingBlocks,
+        cost,
+        toBreak
+      )
+    )
```
Karang, line 276: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+    neighbors.push(
+      new Move(
+        landingBlock.position.x,
+        landingBlock.position.y + 1,
+        landingBlock.position.z,
+        node.remainingBlocks - toPlace.length,
+        cost,
+        toBreak,
+        toPlace
+      )
```
Karang, line 319: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+    )
```
Karang, line 348: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
-    toPlace.push({ x: node.x, y: node.y - 1, z: node.z, dx: 0, dy: 1, dz: 0, jump: true })
+    toPlace.push({
+      x: node.x,
+      y: node.y - 1,
+      z: node.z,
+      dx: 0,
+      dy: 1,
+      dz: 0,
+      jump: true
+    })
```
Karang, line 372: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+      new Move(
+        node.x,
+        node.y + 1,
+        node.z,
+        node.remainingBlocks - toPlace.length,
+        cost,
+        toBreak,
+        toPlace
+      )
+    )
```
Karang, line 385: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+          blockC.position.y,
+          blockC.position.z,
+          node.remainingBlocks,
+          1,
+          [],
+          [],
+          true
+        )
+      )
+    } else if (safeBlockB.safe && safeBlockC.physical) {
```
Karang, line 425: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+          blockB.position.y,
+          blockB.position.z,
+          node.remainingBlocks,
+          1,
+          [],
+          [],
+          true
+        )
+      )
+    } else if (safeBlockC.safe && safeBlockD.safe) {
```
Karang, line 447: No need to make it spawn multiple lines.

lib/movements.js
```diff
@@ -104,213 +102,365 @@ class Movements {
...
+          blockE.position.x,
+          blockE.position.y,
+          blockE.position.z,
+          node.remainingBlocks,
+          1,
+          [],
+          [],
+          true
+        )
+      )
```
Karang, line 463: No need to make it spawn multiple lines.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mineflayer-pathfinder",
-  "version": "1.0.12",
+  "version": "2.0.0",
```
Karang, line 3: Not a major version + version should only be increased in dedicated commit.

readme.md
```diff
@@ -49,27 +47,29 @@ bot.once('spawn', () => {
...
- * Can swim
+
+- Optimized and modernized A\* pathfinding
+- Complexe goals can be specified (inspired by [baritone goals](https://github.com/cabaletta/baritone/blob/master/FEATURES.md#goals) )
+- Customizable movements generator
+- Each movement can have a different cost
+- Can break/place blocks as part of its deplacement
+- Automatically update path when environment change
+- Long distance paths
+- Can swim
```
Karang, line 63: Not needed.

examples/bench.js
```diff
@@ -21,15 +20,29 @@ bot.loadPlugin(pathfinder)
...
   const mcData = require('minecraft-data')(bot.version)
 
   const defaultMove = new Movements(bot, mcData)
-  const goal = new GoalXZ(bot.entity.position.x + 100, bot.entity.position.z)
-  bot.pathfinder.getPathTo(defaultMove, goal, (results) => {
-    console.log('I can get there in ' + results.path.length + ' moves. Computation took ' + results.time.toFixed(2) + ' ms.')
-    bot.quit()
-    process.exit()
-  }, 10000)
+  const goal = new goals.GoalXZ(
```
Karang, line 30: I prefer to import explicitly which goals get used and use `new GoalXZ`

Karang (changes requested):
Can you apply the following changes? It will make the diff clearer and easier to review.

Hopefully, we don't have an 80 character limit in modern IDE, so there is plenty of horizontal space of the screen we can use to avoid scrolling up/down for ages ;)

PR #67, opened 2020-11-06
feat: expose movements and goal

### LINKED ISSUE #57: Expose movements and goal variables
It would be useful to have a getter for the current movements and goal config, alongside the setters.

This is helpful in situations where you want to assign a temporary movement config or goal for a period of time and then revert back to the previous state.

NotSugden:
This PR adds 2 new properties `movements` and `goal`, these are read-only getters

marking as draft until #66 is merged then these can be added to the new typings

resolves #57

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
-/node_modules
+node_modules
 package-lock.json
 yarn.lock
+.vscode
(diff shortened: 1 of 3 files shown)


NotSugden:
@Karang PR is ready

PR #55, opened 2020-10-26
API in the readme

MrAlexEsisteGia:
A quick and dirty wiki to make the plugin much easier to use for beginnners

PR #72, opened 2020-11-19
Set every non diggable block automatically from mcdata

MrAlexEsisteGia:
This should theoretically work but I haven't tested it yet in action

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -29,12 +29,13 @@ class Movements {
     this.allowParkour = true
 
     this.blocksCantBreak = new Set()
-  mcData.blocksArray.forEach(block => {
-  if (block.diggable) return
-  this.blocksCantBreak.add(block.id)
-})
     this.blocksCantBreak.add(mcData.blocksByName.chest.id)
-
+    
+    mcData.blocksArray.forEach(block => {
+    if (block.diggable) return
+    this.blocksCantBreak.add(block.id)
+    })
+    
     this.blocksToAvoid = new Set()
     this.blocksToAvoid.add(mcData.blocksByName.fire.id)
     this.blocksToAvoid.add(mcData.blocksByName.wheat.id)
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -29,7 +29,10 @@ class Movements {
     this.allowParkour = true
 
     this.blocksCantBreak = new Set()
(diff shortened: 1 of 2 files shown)


NotSugden (changes requested):
needs spacing improvements, also might it be better to have a setting (boolean) to go after unbreakable blocks rather than add them all to this `Set` of blocks

NotSugden:
the branch you made for this PR needs rebasing, as it is based off the same branch as your previous PR #55

MrAlexEsisteGia:
> 
> 
> needs spacing improvements, also might it be better to have a setting (boolean) to go after unbreakable blocks rather than add them all to this `Set` of blocks

to me, it seems a bit excessive, it was already marking a few unbreakable blocks manually

TheDudeFromCI (changes requested):
Since we have chests in this list by default, it might be a good idea to add all other inventory holding blocks as well.

MrAlexEsisteGia:
> 
> 
> Since we have chests in this list by default, it might be a good idea to add all other inventory holding blocks as well.

there is no easy way to add them automatically so, wouldn't a bool like "dontMineImportantBlocks" be better?
with also other blocks in it (like diamond blocks - doors - ladders - spawners - ecc...)

MrAlexEsisteGia:
@Karang do I need to change anything or is this good

rom1504:
small note here: if this is merged, it should be merged with "squash and merge" button to avoid having so many commits

Karang:
> @Karang do I need to change anything or is this good

The PR seems fine, can you rebase on master and squash your commit ? For a 4 line change, 20 commits is a bit too much ;)

rom1504:
That was my point about the GitHub squash and merge button, it does the squashing

Karang:
> That was my point about the GitHub squash and merge button, it does the squashing

If I squash it myself it seems like it will change the author. I'd prefer if the author squash it before I merge.

rom1504:
https://github.com/PrismarineJS/mineflayer/commits/master GitHub handles it properly. All these commits with a PR id in parentheses are merged with "squash and merge"
It indicates the original author in addition to having a single clean commit
I think it works pretty well

Karang:
@rom1504  yeah ok it works, it wasn't clear from the gui

PR #77, opened 2020-11-27
Updated so that it avoids gravel and bedrock

DrMoraschi:
I thought the bot would avoid bedrock blocks as it's already in the blocksCantBreak set, but I doesn't seem like that. I thought it's better to avoid bedrock because it's unbreakable, so there's no point in pathfinding through it. Also the bot sometimes dies in falling gravel, so I think it's better to avoid gravel, but I'm not sure, some people may want to pathfind to gravel blocks, so I'm not sure about this PR. There is no need to merge it if you don't want.

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -36,6 +36,8 @@ class Movements {
     this.blocksToAvoid.add(mcData.blocksByName.fire.id)
     this.blocksToAvoid.add(mcData.blocksByName.wheat.id)
     this.blocksToAvoid.add(mcData.blocksByName.lava.id)
+    this.blocksToAvoid.add(mcData.blocksByName.gravel.id)
+    this.blocksToAvoid.add(mcData.blocksByName.bedrock.id)
 
     this.liquids = new Set()
     this.liquids.add(mcData.blocksByName.water.id)


Karang:
Thanks for this.
I think you meant to add those blocks to blocksCantBreak array? because blocksToAvoid is the set of blocks the bot cant walk into (but as bedrock and gravel are both physical blocks, it can walk into them anyway.
For bedrock, it's already not considered for breaking in survival because the breaktime is near infinity https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L120
For gravel, I think it should be better to process it like the flowing water: https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L99
You can add a condition to check if the block above the block you want to break is gravel or sand (and make it configurable in the movement generator:
```js
if (this.dontMineUnderFallingBlock) {
      // false if next to liquid
      if (this.getBlock(block.position, 0, 1, 0).canFall) return false
}
```
You can add the canFall property to the block like its done for the liquids: https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L86 (only adding gravel and sand for now)

DrMoraschi:
I'm sorry I'm don't have enough knowledge in this to do what you have asked for. Could you do it?

Karang:
Take it as a learning opportunity :)
Study a bit the code for the liquid:
https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L40
https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L86
https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L99
And don't worry, the code will be reviewed ;)

DrMoraschi:
Thx

PR #84, opened 2020-12-11
Add ability to navigate through open wooden doors.

nicholasjackson:
Hi,

This little PR adds the ability to navigate through open wooden doors.  If this is acceptable we were also looking at how the bot could open a door that is closed.

Kind regards,

Nic

Co-Authored-By: Erik Veld <erik.veld@gmail.com>

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -15,6 +15,44 @@ const diagonalDirections = [
   { x: 1, z: -1 },
   { x: 1, z: 1 }
 ]
+const doorClosedMetadata = [
+  3,  // closed top left south
+  7,  // closed top right south
+  11, // closed bottom left south
+  15, // closed bottom right south
+  19, // closed top left north
+  23, // closed top right north
+  27, // closed bottom left north
+  31, // closed bottom right north
+  35, // closed top left east
+  39, // closed top right east
+  43, // closed bottom left east
+  47, // closed bottom right east
+  51, // closed top left west
+  55, // closed top right west
+  59, // closed bottom left west
+  63, // closed bottom right west
+]
+
+const doorOpenMetadata = [
+  1,  // open top left south
+  5,  // open top right south
+  9,  // open bottom left south
+  13, // open bottom right south
+  17, // open top left north
+  21, // open top right north
+  25, // open bottom left north
+  29, // open bottom right north
+  33, // open top left east
+  37, // open top right east
+  41, // open bottom left east
+  45, // open bottom right east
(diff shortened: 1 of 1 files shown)


lib/movements.js
```diff
@@ -92,6 +135,15 @@ class Movements {
     for (const shape of b.shapes) {
       b.height = Math.max(b.height, pos.y + dy + shape[4])
     }
+
+    // check if open door, if so treat it like air so the bot can navigate
+    if (b.name.includes('door') && b.material === 'wood' && doorOpenMetadata.includes(b.metadata)) {
```
extremeheat, line 140: Seems like this door function should be written elsewhere, like maybe prismarine-block or prismarine-world, so it can be used within mineflayer also. The hard coded data values seem version specific, so adding this logic elsewhere and exposing it as a property on block might be a good idea.

Karang: The data for the door is already in `block.getProperties()`, you can see the keys available here: https://minecraft.gamepedia.com/Door#Block_data

Also, this code doesn't seem to take into account the direction of the block: if you travel N-S or E-W, the same door could appear open or closed.

TheDudeFromCI, line 140: Also worth noting that since the open door still has a collision box, there is a chance the bot may run into it from the corner in certain cases by treating it like air. This would need to be tested.

SiebeDW:
> To open a closed door, simply call `bot.activateBlock(block)` on the door. This would be done as the bot is traveling along the path, so you wouldn't need to calculate this at pathfinding time.

Would be nice to have an option to toggle it since some servers prevent interacting with doors.

nicholasjackson:
Hey sorry, will get onto some updates this week, just been a little busy.

Karang:
@nicholasjackson What's the status of this PR ? Are you going to continue ?

Karang:
@nicholasjackson Reopen if you want to work on this again.

SinanAkkoyun:
Is this going to be implemented some time?

nicholasjackson:
So sorry about this, I have not had time to move recently. I will pick this back up in the near future, implementing the comments from this PR and raise a new one.

tacohitbox:
any updates on this? @nicholasjackson (no rush btw)

MrJestrik:
it's still not fixed and the bot is still not opening the doors and still breaking the wall xDDD

TravnikovDev:
Guys, that's super confusing issue is still actual

Issue #70, opened 2020-11-09
Fails at parkour


popey456963:
If I wanted to contribute the ability to "parkour" to this module, would you be able to point me in the correct direction to do so?  What would it take in order to allow the program to make jumps?

Karang:
There are 2 steps for adding special moves:
* Make the pathfinder aware that move is valid (and assign it a cost to compare it to other possible moves)
* Implement a control algorithm to execute that move (or more likely modify the existing control algorithm)

First step is done by generating the move in the getNeighbors function: https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L333

Second step is done here https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/index.js#L327 by adjusting the control states and heading direction.

At the moment, only basic parkour is implemented (1 blocks gaps, towering,..). It would be interesting to add more complex movements (>1 block jumps, fall in water (place bucket), ladders and vines).

popey456963:
So, not sure if this deserves a new issue or is under this catergory, but sometimes I see the bot failing at making a simple single block horizontal jump, e.g:

https://v2.femto.pw/d4wy.mp4

It doesn't always happen, take this similary single block jump from standing:

https://v2.femto.pw/4upj.mp4

Here the bot succeeds on the first jump, but then fails the second:

https://v2.femto.pw/m5iw.mp4

All of these are reproducible failures, given the same environment.

Karang:
Yeah, that says the control algorithm need to be improved. Feel free to play with it.

Issue #44, opened 2020-08-29
Cost for placing blocks


apuly:
I've noticed that the bot prefers pillaring up a wall instead of using a detour over nearby stairs, which is sometimes preferable.
Is there a way of setting a cost for placing blocks, in the same way as setting cost for breaking blocks?

Issue #14, opened 2020-06-24
Implement async A*


Karang:
Right now, the A* implementation is sync, when the computation timeout, we return the path to the best node. While the returned path is generally in the good direction, complexes environments like mazes can be deceiving.

To improve the path, more nodes needs to be explored, so the timeout should be removed. But the computation should not block other plugins like physics that should be updated every tick. Additionally, to keep the responsiveness of the api, a path estimation should be returned as soon as possible so the bot can start walking, while the computation continues. The solution is to allow a fixed amount of time per tick to advance the computation and save the computation context to continue where it left on the next tick.

Things that needs to be considered:
* Computation should restart on environment / goal change

This issue track what has to be done to achieve that goal:
- [x] Make A* a class to store the computation context 
- [ ] Implement free motion and path compression (this is needed in case the part of the path that was already walked is updated)
- [ ] Path merging logic
- [ ] Update movement monitoring

rom1504:
That's a really cool feature

TheDudeFromCI:
The implementation is pretty simple and straightforward. (Example is shown [here](https://github.com/TheDudeFromCI/mineflayer-pathfinder/blob/async-path-generation/lib/pathbuilder.js).)

All you need to do is wrap the function inside an object which stores the open and closed sets. Then have a function which only runs a single iteration. Lastly, add a second function to run as many iterations as possible for a given amount of time. This allows you to configure how long you want to run for, or if you want to run indefinitely until a path if found. Or simple want to configure your own kill-switch, sort of speak.

Karang:
> The implementation is pretty simple and straightforward. (Example is shown [here](https://github.com/TheDudeFromCI/mineflayer-pathfinder/blob/async-path-generation/lib/pathbuilder.js).)
> 
> All you need to do is wrap the function inside an object which stores the open and closed sets. Then have a function which only runs a single iteration. Lastly, add a second function to run as many iterations as possible for a given amount of time. This allows you to configure how long you want to run for, or if you want to run indefinitely until a path if found. Or simple want to configure your own kill-switch, sort of speak.

I'm not sure that your rewritting of astar is working (seems to be missing some bits) and it doesn't seem very efficient (why transforming a O(1) lookup into a O(n) https://github.com/TheDudeFromCI/mineflayer-pathfinder/blob/async-path-generation/lib/pathbuilder.js#L41 ?)

[...]

PR #85, opened 2020-12-11
Replace control logic by physic simulation, sprint and advanced parkour

### LINKED ISSUE #80: Sprint usage
An option should allow the bot to use sprint to go to its destination.

Karang:
Inspired by https://github.com/mat-1/mineflayer-baritone with some simplifications.

Closes #80, #70, #14, #44

### DIFF
--- a/examples/test.js
+++ b/examples/test.js
@@ -25,7 +25,7 @@ bot.once('spawn', () => {
 
   bot.on('path_update', (r) => {
     const nodesPerTick = (r.visitedNodes * 50 / r.time).toFixed(2)
-    console.log(`I can get there in ${r.path.length} moves. Computation took ${r.time.toFixed(2)} ms (${nodesPerTick} nodes/tick).`)
+    console.log(`I can get there in ${r.path.length} moves. Computation took ${r.time.toFixed(2)} ms (${r.visitedNodes} nodes, ${nodesPerTick} nodes/tick)`)
   })
 
   bot.on('goal_reached', (goal) => {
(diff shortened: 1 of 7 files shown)


rom1504:
Would be really great to have some automated benchmark measuring how fast to go from A to B. That way it would be easy to see the consequence of this kind of PR

Karang:
Yeah testing it manually is not really practical. There are 3 aspects that needs to be tested for this kind of changes:
* computation time (or number of visited nodes for fixed timeout)
* path quality (did we found a path to destination?)
* path execution (can the bot follow the path without falling ?)
And of course, it is better if the path includes weird blocks (water, half blocks, snow layers, fences, etc...)

rom1504:
I think it would work to have tests like in mineflayer with node-minecraft-wrap, to use a flat world and then do a bunch of set block to try various scenario.
Then make the bot do the pathfinding and measure these things

Karang:
Yeah that's the basic idea, but a bunch of setBlocks will not be practical, we'll need a way to paste schematics or load a preconfigured test world. This could also allow more members to contribute to tests by just providing a schematic with the configuration that fails.

rom1504:
Yeah that sounds pretty good. We could do this as part of a package as
suggested by @TheDudeFromCI there
https://github.com/PrismarineJS/mineflayer/issues/1422

rom1504 (approved):
lgtm
if you tested it manually and it works, let's merge ?

Karang:
> lgtm
> if you tested it manually and it works, let's merge ?

I'm still testing edge cases, I want to be sure it works properly. I'll merge it when its ready.

Issue #54, opened 2020-10-23
Bot gets stuck in water if it tries to place a block under itself while Jumping/Swimming


IceTank:
While in water the bot can get stuck if it tries to place a block.
This can happen when the bot is standing in 1 high water and tries to place a block to get up 1 block. In this case the bot tries to place a block under itself while jumping. While not in water the bot is able to jump high enough to place a block under itself but this is not the case in Water. In Water it is impossible to jump higher then the water Block the player or bot is in.  The Bot tries anyway and gets stuck.

Visualisation:
w = Water
o = Water Block the bot is trying to replace with a block while jumping and bot feet position
i = Bot head position  
s = Solid block
```  
  i
wwoww
sssss
```

IceTank:
Posible fix:
Add this line to movements.js after line 251:
```
if (this.getBlock(node, 0, 0, 0).liquid) return
```

SinanAkkoyun:
It still does not work for me!

Karang:
Fixed, update to 1.3.0.

PR #90, opened 2020-12-23
Bot getting stuck in Water fix

IceTank:
Fixes issue #54

The Bot tries to jump up inside the water block to place a Block below it. 
This is not possible with standard Minecraft physics as the player cannot jump high enough out of the water block to place a block below itself.

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -273,6 +273,7 @@ class Movements {
   }
 
   getMoveUp (node, neighbors) {
+    if (this.getBlock(node, 0, 0, 0).liquid) return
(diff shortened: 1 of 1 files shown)


TheDudeFromCI (approved):

Karang:
This move is possible with minecraft physics. The issue is in mineflayer's physic implementation. There is an open issue about it: https://github.com/PrismarineJS/prismarine-physics/issues/28

Karang:
After some testing, it seems that placing a block while jumping in water only works in 1.13+. I'll merge it for now, we'll add the functionality for 1.13+ when we fix p-physics.

PR #93, opened 2021-01-01
Bridge builder

IceTank:
Allows the bot to build bridges with line of sight to the face it places the block against. 
Prevents NoCheatPlus from kicking to bot for illegal block placement.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,6 +24,7 @@ function inject (bot) {
   let placing = false
   let placingBlock = null
   let lastNodeTime = performance.now()
+  let returningPos = null
   const physics = new Physics(bot)
 
   bot.pathfinder = {}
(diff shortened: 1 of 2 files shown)


IceTank:
Building a staircase is broken. The bot should move back onto the block were it came from to prevent it from getting stuck. Or the bot could jump to place the block.

IceTank:
The bot can now move back into a position were it is able to place the next block to build a staircase. This implementation is really hacky tho. But it might inspire better solutions. I think the main problem is that the bot can end up in a situation were it is trying to place a block in a spot that is occupied by itself. And as it has no way to move out of the way it gets stuck. Maybe we could store a home position in the path the bot can return to if it cant place a block.
Or we could also split the getMoveJumpUp function in movements into two parts were on is placing a support block and the other is placing the actual block it needs to go up and forward by one. If the getMoveJumpUp is split that way we could insert a new movement node that brings the bot back into a position were it is able to place the next block.

Karang:
@IceTank I made a PR to p-physics to add support for edge security https://github.com/PrismarineJS/prismarine-physics/pull/31 when its merge, you can experiment with it to see if its a better solution.

IceTank:
I will make a new Pull request when I have something that works

PR #105, opened 2021-01-12
Legit bot bridging

IceTank:
Some anti-cheat plugins don't allow the placing of blocks if the face the block is being placed agains faces await from the current position. This could break the ability for the Bot to build bridges. The bot can now sneak to the edge of the block to place its next block and automatically return to a previous position if the next block to place would be obstructed by the bot.

https://youtu.be/c4Tvc0L0u_0

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,7 @@ function inject (bot) {
   let placing = false
   let placingBlock = null
   let lastNodeTime = performance.now()
+  let returningPos = null
   const physics = new Physics(bot)
 
   bot.pathfinder = {}
@@ -228,10 +229,15 @@ function inject (bot) {
   }
 
   function moveToEdge(block, edge) {
-    //TODO should be replaced with sneak when it is implemented in prismarine-physics
-    const maxMovement = 0.2
     const noneFalloffDistance = .58
-    const targetPos = block.clone().offset(.5, 0, .5).offset(edge.x * noneFalloffDistance, 1, edge.z * noneFalloffDistance)
+    return moveToBlock(block.offset(edge.x * noneFalloffDistance, 1, edge.z * noneFalloffDistance))
+  }
+
+  function moveToBlock(block) {
+    //TODO should be replaced with sneak when it is implemented in prismarine-physics
+    const maxMovement = 0.1
+    const targetPos = block.clone().offset(.5, 0, .5)
+    console.log(targetPos)
     if (bot.entity.position.distanceTo(targetPos) > 0.001) {
       const targetVec = targetPos.clone().subtract(bot.entity.position).normalize()
       if (maxMovement * maxMovement < bot.entity.position.distanceSquared(targetPos)) {
@@ -282,6 +288,11 @@ function inject (bot) {
       bot.emit('path_update', results)
       path = results.path
       astartTimedout = results.status === 'partial'
+    }
+
+    if (returningPos) {
+      if (!moveToBlock(returningPos)) return
+      returningPos = null
     }
 
     if (path.length === 0) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -227,6 +228,68 @@ function inject (bot) {
...
+    const targetPos = block.clone().offset(0.5, 0, 0.5)
+    console.log(targetPos)
+    if (bot.entity.position.distanceTo(targetPos) > 0.001) {
+      const targetVec = targetPos.clone().subtract(bot.entity.position).normalize()
+      if (maxMovement * maxMovement < bot.entity.position.distanceSquared(targetPos)) {
+        targetVec.scale(maxMovement)
+      } else {
+        targetVec.scale(bot.entity.position.distanceTo(targetPos))
+      }
+      bot.entity.position.add(targetVec)
```
Karang, line 287: It would be nice to not edit the position directly.

Karang (commented):
LGTM, can you add an option to disable this feature if people want to ? (It can be enabled by default)

index.js
IceTank: Yes the problem is that without editing the position directly I would have to tell the bot somehow to correct his position by itself and I don't know how to do that exactly. The position is to precise to be adjusted by mineflayer-pathfinder as far as I know.

IceTank:
Sometimes the bot falls of a bridge when it tries to place the last block. It might have something to do with the bot calling fullStop() in index.js when the path length is zero even tho the last placed block has not yet been confirmed as placed by the server.

Karang:
> Sometimes the bot falls of a bridge when it tries to place the last block. It might have something to do with the bot calling fullStop() in index.js when the path length is zero even tho the last placed block has not yet been confirmed as placed by the server.

Yeah this is a known issue of mineflayer that we discussed a few days ago. The vanilla client doesnt have this issue because it place the block no matter what and only cancel it if the server send a different block update. This is not implemented because getting the right block state to place the block is non trivial (for stuff like stairs, signs, etc..)

However, there is a workaround you can try (and it should improve the towering up too):

Just before https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/index.js#L334 add:
```js
bot.world.setBlockStateId(refBlock.position.offset(placingBlock.dx, placingBlock.dy, placingBlock.dz), 1)
```
This will set the block to stone client side before making the call to place the block. So as the physic is simulated by the client, the stone block will support the bot and prevent it from falling.

IceTank:
I changed how the bot approaches the edge and it does not fall off anymore. It is now more reliable too as it is now self centering.

IceTank:
The Block not being placed after the server confirmed the placing might be favorable when building bridges or stairways. In the current implementation the bot could only fall of if it misstepped or something else weird happens. As it crouches until the block has been placed. But having instant block placement would be good for building 1 by 1 pillars.

SinanAkkoyun:
Hey, the current "solution" does not work. The bot is able to stack some blocks up and then gets watchdog banned on hypixel.

Do you guys have any idea why this still happens? Maybe because of rapid head movement? Do you have any verbal info I could try and take to make a hotfix?

Thanks, I was only able to test tower capability for 2 seconds

(I know the risks and this is totally my fault xD I just want to help make it work)

IceTank:
This pull request only 'fixes' a problem with bridging. Building 1x1 towers is still very unreliable with servers that have any ping as the bot only knows that a block has been placed after the server sends a block update.

SinanAkkoyun:
Interesting, so what does the server think/see when the bot towers up 1x1 so that it bans it?

Or rather what does the bot do to get banned exactly? :)

IceTank:
The bot jumps and sends a places block message to the server when it is high enough to place the block. For the bot the block is not placed until the server confirms it so it falls right back onto the block were it jumped off of. The server thinks the bot jumped and then phased back into the block it just placed resulting in rubber banding or the block not being placed. Some anti cheat plugins might report it as not being vanilla.

PR #107, opened 2021-01-20
Make goto async

u9g:


u9g:
might want to hold off on this, as it's probably a better idea to make a new gotoasync() function instead of just rewriting goto() as that would break existing projects

rom1504 (changes requested):
Can you use the same technique we used in mineflayer ?
Ie make it async but use a callbackify function on top so it both returns a promise and take a callback.

lib/goto.js
```diff
@@ -1,47 +1,57 @@
-function error (name, message) {
-  const err = new Error(message)
-  err.name = name
-  return err
-}
+// function error (name, message) {
+//   const err = new Error(message)
+//   err.name = name
+//   return err
+// }
```
Karang, line 5: If this function is not needed, remove it, don't comment.

lib/goto.js
```diff
@@ -1,47 +1,57 @@
...
 
-  function noPathListener (results) {
-    if (results.status !== 'noPath') return
-    cleanup(error('NoPath', 'No path to the goal!'))
-  }
+    function noPathListener (results) {
+      if (results.status !== 'noPath') {
+        // cleanup(error('NoPath', 'No path to the goal!'))
+      }
+    }
```
Karang, line 26: This callback does nothing. It should reject in case the status is noPath.

lib/goto.js
```diff
@@ -1,47 +1,57 @@
...
 
-  function goalChangedListener (newGoal) {
-    if (newGoal === goal) return
-    cleanup(error('GoalChanged', 'The goal was changed before it could be completed!'))
-  }
+    function goalChangedListener (newGoal) {
+      if (newGoal === goal) {
+        // cleanup(error('GoalChanged', 'The goal was changed before it could be completed!'))
+      }
+    }
```
Karang, line 32: if the goal is changed newGoal !== goal, it should reject the promise

Karang (changes requested):

PR #120, opened 2021-02-21
Add block placement and digging events

CyberPatrick:
In some cases it may be useful

### DIFF
--- a/index.js
+++ b/index.js
@@ -367,8 +367,8 @@ function inject (bot) {
             lastNodeTime = performance.now()
             if (err) resetPath()
             digging = false
+            bot.emit('block_digged', block)
           })
-          bot.emit('block_digged', block)
         })
       }
       return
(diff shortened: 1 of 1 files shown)


Karang:
There are already events to achieve that in mineflayer:
https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#diggingcompleted-block
https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#blockupdatex-y-z-oldblock-newblock

If you feel we need more explicit events, those would have to be added to mineflayer, not pathfinder.

CyberPatrick:
How i can find out that the block was placed by the bot?

Karang:
> How i can find out that the block was placed by the bot?

You can open a PR to add an event to mineflayer here https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/place_block.js#L72

PR #149, opened 2021-05-05
Fix lava swimming

etiaro:


Karang:
This change looks good to me. However, can you run the linter to make it pass the test ? (`npm run fix`)

etiaro:
Okay, done

PR #155, opened 2021-05-17
New Feature - Entity detection

olie304:
- Added checks to ensure entities do not obstruct block placements
- Added weights to make paths with entities in the way less favorable
- Fixed various problems with the movement logic

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,6 +71,12 @@ Also, **for now**, there is only the `pathfinder` module, `movements` and `goals
 
 # Functions:
 
+### bot.pathfinder.goto(goal, callback?)
+Returns a Promise. Resolves when the goal is reached. Rejects on error.
+ * `goal` - Goal instance
+ * `callback` `<function>` - (Optional) The callback is passed two arguments `(err, result)`, where error is `null` on success.
+
+
 ### bot.pathfinder.bestHarvestTool(block)
 Returns the best harvest tool in the inventory for the specified block
  * `Returns` - ?


Karang:
Thanks for this, As it is a bit long I'll need some time to review it properly, but from a quick glance it looks fine. Can you run the linter (`npm run fix`) to fix the formatting and make the tests happy ?

olie304:
> Thanks for this, As it is a bit long I'll need some time to review it properly, but from a quick glance it looks fine. Can you run the linter (`npm run fix`) to fix the formatting and make the tests happy ?

Sure thing, sorry i'm new to node stuff and my IDE's settings were all wacked up. Those fixes will be committed momentarily

rom1504:
let's finish it ?

Karang:
I think it is ready but I have absolutely no time to test it. Its a quite large change and should be at least checked for behavior and performances in some scenarios. Wouldn't be an issue if we had proper test, but i got no time to do that either.

SaubereSache:
Any news on this? @Karang

rom1504:
@IceTank do you feel like reviewing this?

olie304:
I might have to redo this now that it is a quite a few commits behind (I probably should have split it up a little more). I do agree that having some unit tests would be useful but, I have no clue on what the best way of doing that is. If what you mean by unit tests is checking to make sure the bot can actually complete the paths it finds, simulating the world seems like it would not cover everything and there would be countless scenarios. Otherwise if you mean to test whether it finds the correct path through a test course then that is much simpler but, I don't know if there is an efficient way to create, load, run and measure so many test courses.

lib/movements.js
```diff
@@ -103,6 +105,48 @@ class Movements {
...
+    return ents
+  }
+
+  // Faster but doesn't work good for block placement checks. Better for the entity cost system where we want to see if the actual entity is there
+  getEntitiesAt (pos, dx, dy, dz) {
+    const ents = []
+    if (this.allowEntityDetection && pos) {
+      const dPos = new Vec3(pos.x + dx, pos.y + dy, pos.z + dz).floored()
+      for (const ent of Object.values(this.bot.entities)) {
+        const entPos = ent.position.offset(ent.velocity.x, ent.onGround * ent.velocity.y, ent.velocity.z)
```
IceTank, line 141: `ent.onGround * ent.velocity.y` is this a typo? onGround should be a boolean right? Even it this might work you should use something like `ent.onGround ? 0 : ent.velocity.y` instead for better readability.

IceTank: Also why is this even taking entity velocity into account? Why not just check for the entity position and update the path when the bot is about to intersect the entity?

IceTank:
Have you tested this performance wise? It looks like the bot while look through ever entity loaded to check if it intersects with a given position whenever getBlock is called. This looks like it will really tank performance especially with many entities around. Maybe try and see if the bot can still move around when there are 500+ entities in render distance.
A better way to handle this would be to make an array that indexes entities by there floored x, y, z coordinates. So you can look up entities similar to how you look up blocks.

There are also some changes that do not directly have anything to do with the change you want to make. For instance in index.js. Idk how @Karang wants to handle this but a cleaner solution would be to make a separate pull request for those changes.

lib/movements.js
olie304: > Also why is this even taking entity velocity into account? Why not just check for the entity position and update the path when the bot is about to intersect the entity?

So the reason I did this was because I thought that by the time the bot has finished running everything, the entities will probably already be in their next position. So instead of error correcting in a future tick it should just estimate their next position ahead of time since it shouldn't cost much more to do so.

olie304:
> Have you tested this performance wise? It looks like the bot while look through ever entity loaded to check if it intersects with a given position whenever getBlock is called. This looks like it will really tank performance especially with many entities around. Maybe try and see if the bot can still move around when there are 500+ entities in render distance. A better way to handle this would be to make an array that indexes entities by there floored x, y, z coordinates. So you can look up entities similar to how you look up blocks.

I have not done any extensive performance testing but, the bot still functioned fine in a standardly generated world. I imagine that it does cost quite a bit though.

Indexing the entities would be ideal but I think there are a few reasons why I didn't do that. One reason is that most entities move quite a bit so you would still need to update the index fairly often. Another reason is that entities take up more than one block worth of space depending on their bounding box which matters on a sub-block level.
It has been a while since I touched this but, I think there were a few entity update events that could be tied into for updating such an index. Could also just update the entities from a chunk update. So the main issue is dealing with the bounding boxes. Probably the best solution is just to find every block the entity occupies and place it under those blocks in the index.

> There are also some changes that do not directly have anything to do with the change you want to make. For instance in index.js. Idk how @Karang wants to handle this but a cleaner solution would be to make a separate pull request for those changes.

And as I said before, I should have split it up a bit more and it is now a little dated. It would probably be better if I just redo everything using the current branch.

IceTank:
Pathfinder and the path calculations are designed to run within one single tick. So you do not expect entities to move around while calculating a path. You could index all entities into an array at the start of the calculations and then use that instead of going through ever entity on every block that has to be checked. I think it is also save to assume that most entities do not move around when calculating the path. So in my opinion it is unnecessary to add the velocity to there position. Even for entities that do move around they wont get very far in one tick. 
If you do want to remake this or make some changes you should also consider adding a system to the movements to check if the current path or the next few path nodes contain entities. You should expect other entities to move around if the bot is also moving around.

olie304:
Continued in #281

PR #160, opened 2021-05-23
Fix bot sometimes randomly getting stuck in snow

mat-1:


Karang:
This doesn't seems like a correct fix. Can you expand a bit on what situation it get stuck in ?

PR #162, opened 2021-05-25
Add safeOrBreak check to getMoveDiagonal

SeanMcCord:
There are some movements generated that violate the rules because they don't check all the blocks the bot will occupy. For example when a bot moves diagonally and down one block it does not check the final block for safety.

Here is a video demo of the issue: https://youtu.be/0xLNvio0cNA

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -284,6 +284,8 @@ class Movements {
       neighbors.push(new Move(blockC.position.x, blockC.position.y, blockC.position.z, node.remainingBlocks, cost, toBreak))
     } else if (this.getBlock(node, dir.x, -2, dir.z).physical || blockD.liquid) {
       if (blockC.liquid) return // dont go underwater
+      cost += this.safeOrBreak(blockD, toBreak)
+      if (cost > 100) return
       neighbors.push(new Move(blockC.position.x, blockC.position.y - 1, blockC.position.z, node.remainingBlocks, cost, toBreak))
     }
   }


lib/movements.js
```diff
@@ -284,6 +284,8 @@ class Movements {
       neighbors.push(new Move(blockC.position.x, blockC.position.y, blockC.position.z, node.remainingBlocks, cost, toBreak))
     } else if (this.getBlock(node, dir.x, -2, dir.z).physical || blockD.liquid) {
       if (blockC.liquid) return // dont go underwater
+      cost += this.safeOrBreak(blockD, toBreak)
```
Karang, line 287: It was intentional to not allow breaking blocks in diagonals. Can you replace this fix by the simpler check `block.safe` ?

SeanMcCord: What was the intent behind not breaking blocks on diagonals? You bet. My only goal for this PR is to prevent damage to the bot.

Karang: The intent is to keep the search tree manageable, breaking and placing makes the tree grow exponentially much faster than when considering only movements. Restricting breaking and placing to the closest cardinal directions makes very little difference in the path results while keeping the number of nodes explored low and avoiding possibly costly occlusion checks (if you wanted to break diagonal blocks or blocks more than 1 block away, additional raycast checks should be performed).

Karang (approved):

PR #166, opened 2021-05-29
[WIP] Pathfinder block place improvements

IceTank:
I made a new pull request as the old one was messy in its commits


IceTank:
The current idea is that the bot can make movements that are will not be blocked by pathReset and instead wait until the action is done and then allow a path reset. This fixes a few issues were the bot could fall off blocks or fail jumps when resetPath is called mid air. There are other improvements like waiting for block place confirmations before attempting to place the same block again. 
If you have any feedback or if you find problems let me know.

u9g:
Will this be finished?

IceTank:
I am not sure about the un- and re-comment scripts. I may just cut them out of the pr. I don't know if the changes here would benefit pathfinder for the extra complexity they add. The goal if the pr is to give pathfinder the option to ignore path resets so it can execute moves that span multiple ticks without being interrupted. But all of this would have to be tested a lot to make sure it actually works and does not cause more problems. 
I may come back to this in the future but I think we can close this pr for now.

PR #167, opened 2021-06-02
[WIP] Add blocks to not walk on

IceTank:
There is an issue with prismarine-physics were walking speed on soulsand is wrong. Adding a feature were the bot does not walk on certain blocks is also usefull for other things. This is different to blocks to avoid as it can still break the block or build ontop of the block. Im not quite sure if this could be done within block.save so any feedback is welcome.


rom1504:
does this work?

IceTank:
I think it does not work at the moment. This pr is a bit obsolete anyway as it will be possible to add blocks to not step on with the exclusion area pr in user code.

PR #168, opened 2021-06-12
Fixed raycasting not considering block face

IceTank:
GoalPlaceBlock does not check if the raycast opperation hit the right block face when looking for reference blocks to place a block against. This is an issue as anti cheat plugins do raycast calculations on black placement and refuse transactions when the block face is not visible.

### DIFF
--- a/lib/goals.js
+++ b/lib/goals.js
@@ -310,7 +310,7 @@ class GoalPlaceBlock extends Goal {
 
       const block = this.world.raycast(headPos, dir.normalize(), this.options.range)
 
-      if (block && block.position.equals(ref)) {
+      if (block && block.position.equals(ref) && block.face === vectorToDirection (face.scaled(-1))) {
         return { face, to, ref }
       }
     }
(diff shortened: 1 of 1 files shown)


lib/goals.js
```diff
@@ -267,6 +267,7 @@ function distanceXZ (dx, dz) {
  * - facing - the direction the player must be facing
  * - facing3D - boolean, facing is 3D (true) or 2D (false)
  * - half - 'top' or 'bottom', the half that must be clicked
+ * - LOS - true or false, should the bot have line of sight off the placement face. Default true.
```
Karang, line 270: This option is confusing because it's not the definition of line of sight. For instance if you set it to false it would still prune faces that are occluded by an other block. I'm fine with this function not being optional, but if you want to make it an option please find a more appropriate name.

Karang (changes requested):

IceTank:
I changed it so if LOS is false it returns the first found face. My intention with the LOS option was to give the bot the ability to place through walls and on faces it cannot possibly see.

Karang (approved):

PR #169, opened 2021-06-17
Update index.js

adrianfuchs:
I was wondering why hoes dont get picked to harvest leaves and noticed those arguments were set to false while you can easily fix it

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ function inject (bot) {
     let bestTool = null
     for (const tool of availableTools) {
       const enchants = (tool && tool.nbt) ? nbt.simplify(tool.nbt).Enchantments : []
-      const digTime = block.digTime(tool ? tool.type : null, false, false, false, enchants, effects)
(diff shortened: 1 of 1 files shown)


rom1504:
lgtm, what do you think @Karang ?

Karang:
Those arguments were set to false because their value doesn't matter for the comparison. All of those parameters affect the computation in the same way, and we are not interested in the absolute value, but in the relative value for the comparison.
I don't see how this relate to hoe not being picked ? This issue is more likely to be a problem in mcData material hardness values for hoes.

adrianfuchs:
> Those arguments were set to false because their value doesn't matter for the comparison. All of those parameters affect the computation in the same way, and we are not interested in the absolute value, but in the relative value for the comparison.
> I don't see how this relate to hoe not being picked ? This issue is more likely to be a problem in mcData material hardness values for hoes.

I know those "fixes" are not related to hoes not being picked, i just saw it and wondered why it was set to false :) them i just "fixed" it because "why not?"

But if "their value doesn't matter for the comparison", them, ok ¯\_(ツ)_/¯

PR #172, opened 2021-06-21
Fix place wrong block as scaffolding

IceTank:
The issue: 
If the last node contains a block to be placed when pathing and you call bot.equip as soon as the goal_reached event is emitted the placed block used as scaffolding can be the wrong block. 

What I tested:
The bot paths so the last block it stands on was placed by placing while jumping. After the goal_reached event has been emitted the bot then places a block by calling bot.equip and bot.placeBlock with a different block to the blocks used in scaffolding. 
In some cases the block placed as scaffolding is the block equiped after goal_reached was emitted. 
In some cases the bot does not place any block as scaffolding but places the block it should after goal_reached is emitted while jumping.

I think this is caused by bot.equip and bot.placeBlock in the monitorMovement function being async. 
Or it might be an edge case with when the bot jump places a block in the last node as it can reach its goal even if it did not place the block it should. 

I dont know if this fixes that issue but it should not equip and place the wrong block.

### DIFF
--- a/index.js
+++ b/index.js
@@ -394,20 +394,20 @@ function inject (bot) {
         bot.setControlState('jump', true)
         canPlace = placingBlock.y + 1 < bot.entity.position.y
       }
+      bot.equip(block, 'hand', () => {})
+      if (!bot.heldItem || bot.heldItem.type !== block.type) return
       if (canPlace) {
-        bot.equip(block, 'hand', function () {
-          const refBlock = bot.blockAt(new Vec3(placingBlock.x, placingBlock.y, placingBlock.z), false)
-          bot.placeBlock(refBlock, new Vec3(placingBlock.dx, placingBlock.dy, placingBlock.dz), function (err) {
-            placing = false
-            lastNodeTime = performance.now()
-            if (err) {
-              resetPath('place_error')
-            } else {
-              // Dont release Sneak if the block placement was not successful
-              if (!err) bot.setControlState('sneak', false)
-              if (bot.pathfinder.LOSWhenPlacingBlocks && placingBlock.returnPos) returningPos = placingBlock.returnPos.clone()
-            }
-          })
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -394,20 +394,20 @@ function inject (bot) {
         bot.setControlState('jump', true)
         canPlace = placingBlock.y + 1 < bot.entity.position.y
       }
+      bot.equip(block, 'hand', () => {})
```
Karang, line 397: equip is async, if you are not using the callback you should await the promise. This change is likely to introduce more bug than it fix.

The problem you are trying to fix is a race condition when the tick loop is called again before the equip is done. The correct way to solve that would be to add a "lock" ie a variable that is checked first when entering the loop and set to true before doing any async action and false after.

```js
if (equipLock) return // we are already equiping something
equipLock = true // take the lock
await bot.equip(block, 'hand')
equipLock = false // done, release the lock
```

An alternative way of solving this is to cancel the first equip if its called again (we do that for a few other functions like dig or consume). But that would require modifications to equip directly.

@rom1504 any opinion ?

IceTank: Yes my idea was that it continues when it detects that the async equip finishes. I think the problem is that the async functions are used at all or at least how they are used at the moment. 
When I was testing I also noticed that the bot might try and place multiple blocks in the same position as it calls the async blockPlacement function whenever monitorMovement is called.
There should maybe be a lock system that waits for async actions to finish. But you have to make it reliable and resetable in case the path gets recalculated.

IceTank: It might be a good idea to think about splitting index.js into smaller files. If index.js is the package entry file but also handles all movement it can get a bit confusing.

IceTank:
Has been resolved by https://github.com/PrismarineJS/mineflayer-pathfinder/pull/197

PR #173, opened 2021-06-21
Fix reference error if this.options.faces is the default value

IceTank:


Karang:
Can you give more details about the issue ? In what situation refBlock is null/undefined ? If the refBlock is in a chunk not loaded, the goal should be aborted, in that case throwing an exception isn't that bad.

IceTank:
The problem is if you call GoalPlaceBlock without the faces option. In that case it uses all posible faces. If it then interates over these faces to get that blocks for them it can encounter air blocks. `this.world.getBlock(ref)` on air blocks returns null or an object without the `shapes` value. Then `refBlock.shapes` throws a reference error.

Karang:
> The problem is if you call GoalPlaceBlock without the faces option. In that case it uses all posible faces. If it then interates over these faces to get that blocks for them it can encounter air blocks. `this.world.getBlock(ref)` on air blocks returns null or an object without the `shapes` value. Then `refBlock.shapes` throws a reference error.
> The placeBlock.js example won't work without this fix. It would have to check all surrounding blocks if they are air or not and then pass valid faces to GoalPlaceBlock.

ok

PR #175, opened 2021-06-23
Function to stop path finding when safe

IceTank:
This function is different to `setGoal(null)` as it tries to stop when the bot reaches a safe position. Due to errors while building and bridging it also stops on errors like a placing errors or path resets due to block updates, chunk loading etc. 
When resetting paths the bot cancels its velocity so setGoal(null) can cause the bot to fall off edges or fail jumps.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,7 @@ function inject (bot) {
   let placingBlock = null
   let lastNodeTime = performance.now()
   let returningPos = null
+  let stopPathing = false
   const physics = new Physics(bot)
 
   bot.pathfinder = {}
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -418,7 +432,12 @@ function inject (bot) {
     let dz = nextPoint.z - p.z
     if (Math.abs(dx) <= 0.35 && Math.abs(dz) <= 0.35 && Math.abs(dy) < 1) {
       // arrived at next point
+      console.log('stop', stopPathing)
```
Karang, line 435: remove console.log

PR #184, opened 2021-07-26
cleanup Pathfinder#goto when Pathfinder#stop is called

GoogleSites:
Calling `Pathfinder#stop` while a path from `Pathfinder#goto` is being travelled causes it to stop indefinitely, making it throw an error when `Pathfinder#goto` is called again.

I've added the `path_stop` listener to `Pathfinder#goto`, which will clean up when called while the path is still being used.

### DIFF
--- a/lib/goto.js
+++ b/lib/goto.js
@@ -56,7 +56,7 @@ function goto (bot, goal) {
       }, 0)
     }
 
-    bot.on('path_stop', pathStopped);
+    bot.on('path_stop', pathStopped)
     bot.on('goal_reached', goalReached)
     bot.on('path_update', noPathListener)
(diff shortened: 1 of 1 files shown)


Karang:
ci-test doesn't passes, run `npm run fix`

lib/goto.js
```diff
@@ -51,6 +56,7 @@ function goto (bot, goal) {
       }, 0)
     }
 
+    bot.on('path_stop', pathStopped);
```
u9g, line 59:
```suggestion
    bot.on('path_stop', pathStopped)
```

rom1504:
looks ok to me, what do you think @Karang ?

IceTank:
LGTM

PR #189, opened 2021-08-10
Bot will send timeout when it gets to the goal in my experience

u9g:


rom1504:
clearly not a good fix

PR #197, opened 2021-09-06
bot sometimes placing wrong block as scaffolding

IceTank:
Should resolve the issue of the bot placing wrong blocks as scaffolding and trying to place blocks at the same time.


IceTank:
I was testing with a goal were the last block was only reachable while jumping and placing. Then one event goal_reached equip a different block then scaffolding items. Without the lock the bot would sometimes place the equipped block instead of the scaffolding block. But it would also sometimes place no block at all. I think there are also other factors that might hinder block placement.

Karang:
LGTM

PR #198, opened 2021-09-09
Add GoalBlockBreak

IceTank:
This goal places the bot in a position were it is able to break a specific block. The block to break only has to be in line of sight and in reach. I added an example to the examples/example.js

### DIFF
--- a/lib/goals.js
+++ b/lib/goals.js
@@ -333,7 +333,7 @@ function distanceXZ (dx, dz) {
   return Math.abs(dx - dz) + Math.min(dx, dz) * Math.SQRT2
 }
 
-function distance(dx, dy, dz) {
+function distance (dx, dy, dz) {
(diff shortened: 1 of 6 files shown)


lib/goals.js
```diff
@@ -282,6 +333,10 @@ function distanceXZ (dx, dz) {
   return Math.abs(dx - dz) + Math.min(dx, dz) * Math.SQRT2
 }
 
+function distance (dx, dy, dz) {
+  return Math.sqrt(dx ^ 2 + dy ^ 2 + dz ^ 2)
```
Karang, line 337: I'm pretty sure this is already defined in vec3 in one form or an other.

IceTank: I changed it to Vec3.distanceTo. We could also replace distanceXZ used in Goals a lot with Vec3.xyDistanceTo

PR #206, opened 2021-09-29
check if a goal is still valid before checking if the goal has changed

u9g:


IceTank:
I don't think every goal has to implement the function `isValid()` it should work if only the methods that override the default method should implement a `isValid()` method.

rom1504:
I will close this next week if no feedback given

IceTank:
I think the pr is fine the isValid() method is removed from every goal. It should be fine if isValid just returns true by default in the Parent Goal class. Only dynamic goals where the goal could get invalid should need the isValid method.

@u9g Can you refactor it? It should be ready to merge then.

rom1504:
If you want, you can directly edit this @IceTank , you have rights to do it

u9g:
> I think the pr is fine the isValid() method is removed from every goal. It should be fine if isValid just returns true by default in the Parent Goal class. Only dynamic goals where the goal could get invalid should need the isValid method.
> 
> @u9g Can you refactor it? It should be ready to merge then.

Done.

lib/goals.js
```diff
@@ -233,6 +239,13 @@ class GoalCompositeAny extends Goal {
     }
     return false
   }
+
+  isValid () {
+    for (const i in this.goals) {
+      if (!this.goals[i].isValid()) return false
+    }
+    return false
```
IceTank, line 247: This should be `return true`

lib/goals.js
```diff
@@ -267,6 +280,13 @@ class GoalCompositeAll extends Goal {
...
```
IceTank, line 288: Same here

IceTank (changes requested):

lib/goals.js
u9g: done!

IceTank (approved):

IceTank:
@Karang anything to add? Or can we merge this?

PR #211, opened 2021-10-16
Add infiniteLiquidDropdownDistance boolean

IceTank:
to movement

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -192,6 +192,7 @@ declare module 'mineflayer-pathfinder' {
 		public scafoldingBlocks: number[];
 
 		public maxDropDown: number;
+		public infiniteLiquidDropdownDistance: boolean;
 		public digCost: number;
 		public placeCost: number;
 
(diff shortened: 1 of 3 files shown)


rom1504:
please explain what this is doing

you can take this comment in general: PR should be composed of a code change *and* an explanation on why you think this is good
reverse engineering your change to try to guess your intention isn't time well spent for the reviewer

SaubereSache:
If `infiniteLiquidDropdownDistance` isn't set, prevents the bot from jumping more than `maxDropDown` into liquids.

rom1504:
@IceTank can you add to doc in readme ?

IceTank:
To Clarify the pull request:
This pull request adds an option to change the behavior of maxDropDown. Pathfinder currently ignores maxDropDown (distance) when the landing block is water (or a liquid and safe). With this new setting changed to false the max falling distance is always maxDropDown distance regardless off the landing block.

About the comment from SaubereSache. I think it should work (?). I have not tested it yet but I will update the pull request when I know if it is working.

PR #216, opened 2021-10-31
Fix diagonal move not considering collision height when jumping up diagonally

IceTank:
Fixes the bot getting stuck in this Scenario:
image
When moving from the diamond block to the gold block it gets stuck as it tries to jump over the wall even tho it cannot make that jump. By stuck I mean not moving at all as the simulations of the jump fail causing the bot to not move at all.

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -239,22 +239,28 @@ class Movements {
     let cost = Math.SQRT2 // move cost
     const toBreak = []
 
-    const blockC = this.getBlock(node, dir.x, 0, dir.z)
+    const blockC = this.getBlock(node, dir.x, 0, dir.z) // Landing block or standing on block when jumping up by 1
     const y = blockC.physical ? 1 : 0
+
+    const block0 = this.getBlock(node, 0, -1, 0)
 
     let cost1 = 0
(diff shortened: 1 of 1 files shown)


lib/movements.js
```diff
@@ -239,22 +239,28 @@ class Movements {
     let cost = Math.SQRT2 // move cost
     const toBreak = []
 
-    const blockC = this.getBlock(node, dir.x, 0, dir.z)
+    const blockC = this.getBlock(node, dir.x, 0, dir.z) // Landing block or standing on block when jumping up by 1
     const y = blockC.physical ? 1 : 0
 
+    const block0 = this.getBlock(node, 0, -1, 0)
```
Karang, line 245: You moved this out of a condition but don't use it elsewhere, why ?

Karang (commented):
LGTM one minor comment

lib/movements.js
IceTank: Its used for the relative height calculation at line 254 and 263

Karang: ok

Karang (approved):

PR #219, opened 2021-11-01
Removed callbacks form function use

IceTank:
Get pathfinder ready for mineflayers callback removal

### DIFF
--- a/index.js
+++ b/index.js
@@ -392,12 +392,15 @@ function inject (bot) {
         const block = bot.blockAt(new Vec3(b.x, b.y, b.z), false)
         const tool = bot.pathfinder.bestHarvestTool(block)
         fullStop()
-        bot.equip(tool, 'hand', function () {
-          bot.dig(block, function (err) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -392,12 +392,15 @@ function inject (bot) {
...
-        bot.equip(tool, 'hand', function () {
-          bot.dig(block, function (err) {
-            lastNodeTime = performance.now()
-            if (err) resetPath('dig_error')
-            digging = false
-          })
+        bot.equip(tool, 'hand').then(function () {
+          bot.dig(block)
+            .catch(_ignoreError => {
+              resetPath('dig_error')
```
rom1504, line 398: you need to do digging=false here or this is a behavior change

IceTank: I think it should work as the .then after it always triggers? The original code makes digging = false after dig ended with an error or not. So this should do the same? Or am I wrong.

rom1504: ah yes true

PR #220, opened 2021-11-01
Exclusion area

IceTank:
Adding exclusion zone support

### DIFF
--- a/examples/exclusionArea.js
+++ b/examples/exclusionArea.js
@@ -79,7 +79,7 @@ bot.once('spawn', () => {
           case 'place':
             bot.pathfinder.movements.exclusionAreasPlace.push(isExcluded)
             break
-        } 
+        }
         // At 5. The bot avoids the area most of the time but can still move into and out of it.
         bot.pathfinder.movements.exclusionAreaPower = 5
         bot.pathfinder.setMovements(bot.pathfinder.movements)
--- a/examples/exclusionArea.js
+++ b/examples/exclusionArea.js
@@ -1,0 +1,104 @@
+const mineflayer = require('mineflayer')
+const { pathfinder, Movements } = require('mineflayer-pathfinder')
+const { GoalNear, GoalBlock, GoalXZ, GoalY, GoalFollow } = require('mineflayer-pathfinder').goals
+
+if (process.argv.length > 6) {
+  console.log('Usage : node example.js [<host>] [<port>] [<name>] [<password>]')
+  process.exit(1)
+}
+
+const bot = mineflayer.createBot({
+  host: process.argv[2] || 'localhost',
+  port: parseInt(process.argv[3]) || 25565,
+  username: process.argv[4] || 'exclusionAreaBot',
+  password: process.argv[5]
+})
+
+bot.loadPlugin(pathfinder)
+
+bot.once('spawn', () => {
+  // Once we've spawn, it is safe to access mcData because we know the version
+  const mcData = require('minecraft-data')(bot.version)
+
+  // We create different movement generators for different type of activity
+  const defaultMove = new Movements(bot, mcData)
+  bot.pathfinder.setMovements(defaultMove)
+
+  bot.on('path_update', (r) => {
+    const nodesPerTick = (r.visitedNodes * 50 / r.time).toFixed(2)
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -157,7 +157,7 @@ function inject (bot) {
       }
     }
 
-    if (!bot.pathfinder.enablePathShortcut || path.length === 0) return path
+    if (!bot.pathfinder.enablePathShortcut || stateMovements.exclusionAreas.length === 0 || path.length === 0) return path
```
IceTank, line 160: Does this even do anything?

IceTank:
I think it would be nice to have multiple different types of exclusion zones. Maybe one for placing block one for placing blocks and one for stepping in/on. But I think I would have to rewrite a lot of code in movements.js for that. It seams like the code to check if you can step on something is the same as the code that checks if you can break something. So I would have to separate that first.

IceTank:
I updated it so it now includes different exclusion area checks for placing / breaking blocks and stepping into blocks.

rom1504:
can you add something to doc ? (readme)

IceTank:
I think I want to change it so exclusion area functions do not return booleans but numbers. Basically if the returned number is higher then 100 it will always exclude that block from path calculations. But if the value is only slightly higher then 0 it will only add weight to that specific node. With that change you could also implement other stuff like slowdown by soul sand or trying to avoiding mobs when possible.

IceTank:
Ok I updated it so each function returns a number to act as extra weight when calculating the path. If you want to exclude a specific block at all you would return 100 for that specific block. If you just want to make pathfinder avoid that block with a higher cost you would return a smaller cost.
Using functions like this could also make it possible to combine checks like blocksToAvoid, blocksCantBreak or doNotCreateFlow into one thing bigger function. So you don not have to check for every excluded block and block restriction in every movement function.

examples/exclusionArea.js
```diff
@@ -56,6 +56,8 @@ bot.once('spawn', () => {
           bot.chat('I can\'t see you')
           return
         }
+        const type = cmd[3].trim()
```
ratuspro, line 79: Should be cmd[2]

examples/exclusionArea.js
```diff
@@ -56,6 +56,8 @@ bot.once('spawn', () => {
...
+        if (!['break', 'step', 'place'].includes(type.toLowerCase())) return bot.chat('type must be "break", "step" or "place"')
         const radius = Number(cmd[2])
```
ratuspro, line 83: Should be cmd[3]

ratuspro:
Hey @IceTank, we have briefly discussed some features that could be added to this one on Discord.
I would be more than happy to contribute to this PR. I already found a couple of small bugs with some indices.

Also, I recall you saying on Discord that you were not very pleased with the approach followed. 
Do you intend to do some major changes?

IceTank:
> Also, I recall you saying on Discord that you were not very pleased with the approach followed.
Do you intend to do some major changes?

I already changed what I wanted to change about the pull request with [this](https://github.com/PrismarineJS/mineflayer-pathfinder/pull/220/commits/4ba39d099c9b669a18dc3f4a24d30de5d4206cfe) commit. I changed the exclusion area functions not returning boolean but numbers instead. So that the exclusion functions contribute to a given node cost instead of making the node inaccessible.

ratuspro:
Ok!
I'll do some testing later today to see if everything is running adequately and perhaps add an example to avoid stepping on blocks by block type.

IceTank:
I don't know how Karang would want to handle it but this system could also replace all the different checks for not stepping into other hazardous stuff. There might be an argument that simple if checks are more efficient then having to call functions tho.

Karang:
> I don't know how Karang would want to handle it but this system could also replace all the different checks for not stepping into other hazardous stuff. There might be an argument that simple if checks are more efficient then having to call functions tho.

The easiest way to answer that question is to benchmark it. Anyway, this don't have to be done in this PR. We could add the exclusionArea feature, see how it behave, and if we feel confident about it, remove the other checks in the future.

rom1504:
what's the status of this PR?

IceTank:
Should be ready to merge. Some extra testing would be nice but I think it is working from the testing I have done.

IceTank:
I am not 100% happy with the name tho. It is not really a exclusion area but more off a cost function modifier system. You can do a lot more then just defining exclusion areas.

PR #235, opened 2022-01-06
Fix uncaught exception when bestHarvestTool returns null

IceTank:


SzczurekYT:
Here it is. Tanks, and hope it will be merged soon.

Hedgie1:
I was just looking for what was causing error in my code and found this, lol.

SzczurekYT:
Cool, thanks for the fix. @IceTank

PR #240, opened 2022-01-20
Indicate path stopped to caller

SaubereSache:
pathfinder#goto now throws if stopped. 
The path hasn't been completed, and as such, the call should not indicate success.


rom1504:
did you check this doesn't break anything?
how did you test it?

SaubereSache:
I've been using this change in my bot for some time, it does not seem to break anything.

PR #243, opened 2022-02-07
fix default scaffolding blocks

IceTank:


lib/movements.js
```diff
@@ -70,8 +70,8 @@ class Movements {
     this.replaceables.add(mcData.blocksByName.lava.id)
 
     this.scafoldingBlocks = []
-    this.scafoldingBlocks.push(mcData.blocksByName.dirt.id)
-    this.scafoldingBlocks.push(mcData.blocksByName.cobblestone.id)
+    this.scafoldingBlocks.push(mcData.itemsByName.dirt.id)
+    this.scafoldingBlocks.push(mcData.itemsByName.cobblestone.id)
```
rom1504, line 74: Scaffolding blocks are items?

Karang: yes, the name is misleading, but this data is used to select the block item in the inventory
this isn't a bug properly speaking because for the examples (dirt and cobblestone) the item id = the block id
however, it has been confusing people for a long time when trying to add other blocks where item id != block id

Karang (approved):

lib/movements.js
IceTank: For versions below 1.16 (?) cobblestone and dirt had the same ids as blocks and as items. But that changed in newer versions so the bot was unable to use dirt and cobblestone as scaffolding blocks by default.

rom1504:
in another PR it would be good to rename this.scafoldingBlocks

Issue #147, opened 2021-04-28
Add example about using `Movements` class


u9g:

sefirosweb:
Similar to this?

```js

        movements.allowSprinting = false // Disable sprint for avoid loss stamina
        movements.canDig = false // disable destoy blocks
        movements.blocksToAvoid.delete(mcData.blocksByName.wheat.id) // Enable walk over wheat
        movements.blocksToAvoid.add(mcData.blocksByName.sweet_berry_bush.id) // Disable walk over sweat
        movements.blocksCantBreak.add(mcData.blocksByName.sweet_berry_bush.id) // Disablle destroy sweet_berry
        movements.blocksCantBreak.add(mcData.blocksByName.cactus.id) // Disablle destroy sweet_berry

```

PR #244, opened 2022-02-07
Creating an example for Movements class.

Zn10plays:
Boilerplate working on issue #147

### DIFF
--- a/examples/movements.js
+++ b/examples/movements.js
@@ -1,0 +1,33 @@
+
+const mineflayer = require('mineflayer')
+const { pathfinder, Movements } = require('mineflayer-pathfinder').pathfinder
+const { GoalNear } = require('mineflayer-pathfinder').goals
+
+const bot = mineflayer.createBot({ username: 'Player' })
+
+bot.loadPlugin(pathfinder)
+
+bot.once('spawn', () => {
+
+  const mcData = require('minecraft-data')(bot.version)
+  const defaultMove = new Movements(bot, mcData)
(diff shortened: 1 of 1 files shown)


Zn10plays:
Anyone who why its failing.

IceTank:
The example does not work as you are importing pathfinder in a wrong way. It should be `const { pathfinder, Movements } = require('mineflayer-pathfinder')` instead of `const { pathfinder, Movements } = require('mineflayer-pathfinder').pathfinder`. You should also add some extra info about what this example does and change the createBot to be able to join different servers that are running on none standard ports. Look at [this](https://github.com/PrismarineJS/mineflayer/blob/master/examples/chatterbox.js) example for an example.

examples/movements.js
```diff
@@ -0,0 +1,42 @@
...
+  customMoves.blocksToAvoid.add(mcData.blocksByName.carrot.id)
+
+  // To initialize the new movements use the .setMovements method.
+  bot.pathfinder.setMovements(customMoves)
+
+  bot.on('chat', function (username, message) {
+    if (username === bot.username) return
+
+    if (message === 'come') {
+      const target = bot.players[username] ? bot.players[username].entity : null
```
IceTank, line 32: This can be simplified to `const target = bot.players[username]?.entity`

Zn10plays: resolved!

Zn10plays:
Finished! And again ready for review

PR #248, opened 2022-02-19
Banish callbacks to the shadow realm

IceTank:


rom1504:
yeah let's do that
this should prob go in a 2.0.0 though

do you know if many people use callbacks with pathfinder?

u9g (approved):
lgtm

IceTank:
> do you know if many people use callbacks with pathfinder?

I am not sure. But there was not really anything besides goto that used callbacks. Goto is a somewhat newer feature. So I don't think any older plugins use it.

PR #249, opened 2022-02-22
Refactor GoalLookAtBlock and depricate GoalBreakBlock

IceTank:
This should make it less confusing when digging and placing blocks. As now both GoalPlaceBlock and GoalLookAtBlock take similar augments.
I also updated the example for placing and breaking blocks. It now includes an example off how to mine blocks.


Zn10plays (approved):
LGTM

PR #250, opened 2022-03-08
Release 1.10.0

IceTank:
Release 1.9.2

### DIFF
--- a/history.md
+++ b/history.md
@@ -1,6 +1,6 @@
 # History
 
-# 1.9.2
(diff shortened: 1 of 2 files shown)


history.md
```diff
@@ -1,5 +1,25 @@
 # History
 
+# 1.9.2
```
rom1504, line 3: lot of new features so this should be 1.10.0

IceTank: Done

PR #255, opened 2022-03-25
Add getPathFromTo

IceTank:
Exposes a way to have pathfinder calculate a path from any given point to a given goal. Including path optimization and starting position abstraction.


rom1504:
lgtm
I would appreciate a review from @Karang

Karang (approved):
LGTM

PR #267, opened 2022-05-28
Fix obstructed diagonal movement

BattleDash:
When moving diagonally, if there are blocks obstructing the movement, the bot will get stuck forever.

In this example, the red block is the target, the blue block is where the bot is, and the yellow blocks are obstructing the diagonal movement.
javaw_riiJZcaPA9

This checks if there are obstructing blocks, and moves around them.


BattleDash:
[Identified root cause](https://github.com/PrismarineJS/prismarine-physics/issues/67), turns out this worked on versions lower than 1.18.

PR #274, opened 2022-06-24
Force looking at block when digging

IceTank:


IceTank:
This will help when the bot is flagged by anti cheat for not looking at a block when starting to dig a block

PR #276, opened 2022-07-01
Make something better

liulihaocai:


liulihaocai:
> > Make something better
> 
> Can you elaborate? The first two changes make sense but why did you remove the liquid check? Is there a case where this check breaks something?

because in line `373`, if `blockC.liquid == true` will go into another if-else block, then the blockC.liquid must be false

PR #277, opened 2022-07-03
update npm packages

dement6d:


rom1504:
this shouldn't have been merged @IceTank 
we don't need to update the packages thanks to ^ ; read about it at https://stackoverflow.com/a/22345808

IceTank:
> this shouldn't have been merged @IceTank we don't need to update the packages thanks to ^ ; read about it at https://stackoverflow.com/a/22345808

I know what the `^` does. But not everyone contributing to the project always bumps the dependencies when using new features from other modules and instead expect all dependencies to be up to date. A lot off issues on the discord help channel come from outdated dependencies that have not been updated as they still match the current required dependencies. This is indeed not necessary when you are installing pathfinder fresh in a new project but old projects can break due to untraceable dependency errors. I had that happen before. So imo this should be done every once in a while to ensure all valid installations work. This should have been done when releasing pathfinder 2.x but I forgot.

PR #281, opened 2022-07-20
Entity Avoidance Feature

olie304:
Continuation of #155
As requested, I made it so that the entities/intersected blocks are indexed before computing each path. The way it passes the indexed data to the Movements object in use is kind of ugly but, just using what already exists I couldn't think of a better way to do it.

I have tested this with 500+ entities and it seems to handle it no-problem but more extensive performance testing may be necessary.

I'm working on a basic internal test for checking that the feature is functional


olie304:
The unit tests are done. I also did some performance testing and it takes ~10ms to index and path with 10,000 entities present, so I think it will be alright.

lib/movements.js
```diff
@@ -478,6 +528,11 @@ class Movements {
 
     if (this.getBlock(node, 0, 0, 0).liquid) return // cant jump from water
 
+    let cost = 1 // Mostly used for checking if entities are in the jump path. Will be useful incase we want to prefer a path with no potential hostiles that can knock us out of our jump
```
IceTank, line 531: Should cost not start at 0? Cost used to be defined like this `const cost = this.exclusionStep(blockB) + 1` before being added to the move. Now it would be one more.

index.js
```diff
@@ -154,6 +165,37 @@ function inject (bot) {
...
+        for (let y = minY; y < maxY; y++) {
+          entIntersections[y] = entIntersections[y] ?? {}
+          for (let x = minX; x < maxX; x++) {
+            entIntersections[y][x] = entIntersections[y][x] ?? {}
+            for (let z = minZ; z < maxZ; z++) {
+              entIntersections[y][x][z] = entIntersections[y][x][z] ?? 0
+              entIntersections[y][x][z]++ // More ents = more weight
+            }
+          }
+        }
```
IceTank, line 194: Have you tried it with just indexing the object like so ```entIntersections[`${x}${y}${z}`]````this might actually be faster then having to index nested objects.

index.js
```diff
@@ -69,6 +71,7 @@ function inject (bot) {
 
   bot.pathfinder.getPathFromTo = function * (movements, startPos, goal, options = {}) {
     const optimizePath = options.optimizePath ?? true
+    const resetEntIntersects = options.resetEntIntersects ?? true
```
IceTank, line 74: Looks like this is missing from the type definition

lib/movements.js
olie304: The initial cost should be 1 however, since I changed the function to use a shared cost variable, the '+ 1' will need to be removed from each exclusionStep()

index.js
olie304: Agreed, that is a lot neater. Some surface level tests also say that it is ~10% faster

index.js
```diff
@@ -11,6 +11,8 @@ const Vec3 = require('vec3').Vec3
 const Physics = require('./lib/physics')
 const nbt = require('prismarine-nbt')
 
+const passableEnts = require('./lib/passableEnts.json')
```
IceTank, line 14: Is there a reason to make this const? I see some off the entities are things like fireballs or ender dragon balls. Some plugins or user code could make pathfinder avoid those too if the list is editable.

index.js
```diff
@@ -81,6 +84,12 @@ function inject (bot) {
       const b = bot.blockAt(p)
       start = new Move(p.x, p.y + (b && dy > 0.001 && bot.entity.onGround && b.type !== 0 ? 1 : 0), p.z, movements.countScaffoldingItems(), 0)
     }
+    if (movements.allowEntityDetection) {
+      if (resetEntIntersects) {
+        stateMovements.entIntersections = {}
+      }
+      updateCollisionIndex(movements.entIntersections)
```
IceTank, line 91: Why is entIntersections not always cleared when updateCollisionIndex is called? updateCollisionIndex does not seam to ever remove or decrease any entries?

IceTank: Or is there some other place where entIntersections is cleared?

IceTank: Ah ok so entIntersections is cleared on every path reset. So does this mean that the entity map is always recalculated on path calculations? Why the resetEntIntersects check then?

lib/movements.js
```diff
@@ -26,13 +26,15 @@ class Movements {
...
     this.liquidCost = 1
+    this.entityCost = 1
 
     this.dontCreateFlow = true
+    this.dontMineUnderFallingBlock = true
     this.allow1by1towers = true
     this.allowFreeMotion = false
     this.allowParkour = true
     this.allowSprinting = true
-    this.dontMineUnderFallingBlock = true
```
IceTank, line 35: Is the dontMineUnderFallingBlock option removed complete now? If so this would be a breaking change. Can you back port this option for the new system?

index.js
olie304: The main motivation for the `resetEntIntersects` option is so that a user can append their own index if they wanted to without it being cleared when the path is calculated. I am not too sure when this would be useful outside of tests.

lib/movements.js
olie304: It is still there, I just moved it up to the top so that it is next to `this.dontCreateFlow`

index.js
olie304: Implemented. I made the `passableEnts` as well as `interactableBlocks` into public fields on Movements. I also added another field `entsToAvoid` which will cause the indexer to add the max cost to any blocks containing those entities. I figure these could be useful for anyone trying to add their own legacy/modded additions to custom movement types.

IceTank:
On small thing. Can you rename the things like ents to entities? I say the confusion caused by having to think about what this variable does is not outweigh by the few characters you save. Plus all other variables are named in full length too these would be the only one doing it differently. For variables that are used in closed scope this is not that big off an issue. But public API facing properties should be kept consistent and readable.

IceTank:
Also another thing I just noticed. There is API documentation in the readme. Can you add the changes you made to the code to the API documentation? Should be ready to be merged then.

PR #291, opened 2022-09-03
GoalLookAtBlock optimization

Saiv46:
If block is out of reach - it'll behave as `GoalNear`, avoiding expensive computing.


IceTank:
This does not optimize anything. GoalLookAtBlock already acts like goalnear when out off reach.

Issue #129, opened 2021-03-07
Pathfinde don't want walk "over" wheat


sefirosweb:
Hello I saw when pathfinder want to cross wheat they think is a solid block (or similar as lava) and try to go with other way

Where is saved this info for made a PR?

Karang:
This is a configurable option from user code, by default pathfinder dont want to walk into crops (to avoid messing farms) https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L45

sefirosweb:
Oooh i see,

Then i need to use in movements:

```js
 movements.blocksToAvoid.delete(XXX)
```

But this not have a sense, only affects to wheat, in this case must be add potatoes, carrots, melons...
Well al plants

Right?

For my case I can remove it, but for generic use maybe need to best way is add a "avoid plants"

Karang:
The id that are given must be the id of blocks, we cant give categories. But you are right this doesnt work for other plants. Maybe we shouldn't put any default and let the user add them for their needs.

sefirosweb:
I'm align, maybe only need to document this behavior if anyone want to use it,

In my case i'm making a farmer bot, and when is a full of wheat they can't walk he gets totally locked up xD

magicaltoast:
I think that would be sane default to allow walking on crops but now 'falling' on them because only this breaks them, and add to constructor something like allow falling on crops

sefirosweb:
Guys I close the issue because this can be changed easly for each one

```js
        movements.blocksToAvoid.delete(mcData.blocksByName.wheat.id)
        movements.blocksToAvoid.add(mcData.blocksByName.sweet_berry_bush.id)
```

Also it is not a "problem" it is rather subjective for the purpose of your bot

maximmasiutin:
@sefirosweb - can you please check that after #303 is is OK by default, without the need to add lines?

Issue #102, opened 2021-01-09
Wheat cannot be replaced when placing block (unlike grass)


koneshnoVasya:
seems like a bot treats a wheat like a grass (usually you can replace grass with another block, but not wheat)

https://user-images.githubusercontent.com/42637020/104086675-efe97e80-526a-11eb-877c-b1e18c26f55f.mp4

Karang:
Solved. Update to 1.3.3

koneshnoVasya:
thanks

PR #303, opened 2022-12-04
Wheat is not an unbreakable block and should not be avoided - but added misleadingly to the two lists

maximmasiutin:
Wheat is not an unbreakable block and should not be avoided. There are other crops similar to wheat, such as potatoes or carrots. Farmland (not wheat) may be harmed by jumping rather than by moving. However, categorizing wheat as an unbreakable item is misleading, as it does not require breaking it to pass, moreover, it can be broken if needed. There were issues reported as #102 and #129, and the special handling of wheat (rather than carrot and potato) is still misleading. This issue can be solved in each of the applications by movements.blocksToAvoid.delete(wheat) and movements.blocksCantBreak.delete(wheat), but these default settings cannot be justified, they are confusing and not explicit, it was hard to understand from the documentation why they were implemented and why a bot stuck in wheat.

### DIFF
--- a/lib/movements.js
+++ b/lib/movements.js
@@ -41,7 +41,6 @@ class Movements {
 
     this.blocksCantBreak = new Set()
     this.blocksCantBreak.add(registry.blocksByName.chest.id)
-    this.blocksCantBreak.add(registry.blocksByName.wheat.id)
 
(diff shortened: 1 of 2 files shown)


IceTank:
Makes sense Ill merge

PR #311, opened 2023-01-01
Remove mcData param since its not used by the JS

rtm516:


IceTank:
This is a breaking API change and cannot be merged. Changing the number of required arguments will break any existing projects using this feature. What you can do is make the movements option optional instead with ? after the argument.

rtm516:
> Changing the number of required arguments will break any existing projects using this feature.

The JS doesn't take more than 1 argument anyway this TS definition is wrong.
See: https://github.com/PrismarineJS/mineflayer-pathfinder/blob/master/lib/movements.js#L20

PR #312, opened 2023-02-01
Block face position fix

WhoTho:
Same fix I did for mineflayer's visible face math. ([PR](https://github.com/PrismarineJS/mineflayer/pull/2801))

The math for getting the position relative to the center of the block was off by a bit.


IceTank:
LGTM

Issue #313, opened 2023-02-05
Bot trying to open the chest?¿? when he's walking on the chest


sefirosweb:
Hi guys, I saw this error a few times when is walking over interactable blocks,

The bot tries to open the chest when is over him, I think is trying to place block, but is a chest, it need to use sneak, but he did didn't it

The bot stucks clicking to chest a lot of time per second

I attach video: https://youtu.be/7TpLeE_r4oc

I'm using code based on mineflayer-builder, to go position with place block

```ts
    const movements = new Movements(bot, mcData)
    movements.canDig = false
    movements.maxDropDown = 10
    movements.allowSprinting = false
    bot.pathfinder.searchRadius = 10

...
    const { facing, is3D } = build.getFacing(action.state, properties.facing)
    const goal = new goals.GoalPlaceBlock(action.pos, bot.world, {
        faces,
        facing: facing,
        facing3D: is3D,
        half,
        range: placementRange,
        LOS: placementLOS
    })

    bot.pathfinder.tickTimeout = 200
    bot.pathfinder.setMovements(movements)
    await bot.pathfinder.goto(goal)
```

![image](https://user-images.githubusercontent.com/20754836/216847046-38a84ae1-b55f-4ea7-b99f-c4e419fe35a4.png)

IceTank:
Yes intractable block check is not implemented. There is a list of blocks a player can interact with here https://github.com/PrismarineJS/mineflayer-builder/blob/master/lib/interactable.json but I don't have time to implement that. A duct tape fix would be to sneak on every block place.

sefirosweb:
@IceTank  yes i saw!

Right now I'm trying to fix it, but I'm doing tests
https://github.com/sefirosweb/mineflayer-pathfinder/blob/8adbdfe74de18b4a78db6db54b75fd86c888282e/index.js#L497

IceTank:
Fixed by #314

PR #314, opened 2023-02-06
Added "sneak" on interact blocks to avoid open it

sefirosweb:
Add sneak action when bot tries to place block on interactable blocks

Issue related: https://github.com/PrismarineJS/mineflayer-pathfinder/issues/313


IceTank:
LGTM

PR #317, opened 2023-04-08
Update goals.js

GenerelSchwerz:
fix GoalBreakBlock

### DIFF
--- a/lib/goals.js
+++ b/lib/goals.js
@@ -207,13 +207,13 @@ class GoalLookAtBlock extends Goal {
 // Path into a position were a blockface of block at x y z is visible.
 // You'll manually need to break the block. THIS WONT BREAK IT
 class GoalBreakBlock extends Goal {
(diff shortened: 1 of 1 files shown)


lib/goals.js
```diff
@@ -207,13 +207,13 @@ class GoalLookAtBlock extends Goal {
 // Path into a position were a blockface of block at x y z is visible.
 // You'll manually need to break the block. THIS WONT BREAK IT
 class GoalBreakBlock extends Goal {
-  constructor (x, y, z, bot, options = {}) {
+  constructor (pos, bot, options = {}) {
```
IceTank, line 210: Yeah but why? Documentation in readme.md says `GoalBreakBlock(x, y, z, bot, options)`. I can't merge it without a new major release as it would be a breaking API change.

IceTank:
Is end is indeed broken. I can merge when ready.

rom1504:
let's not make a breaking change for such a minor reason

PR #324, opened 2023-06-22
Fixed typo: "physicsTick"

FreezeEngine:


IceTank:
Mineflayer used to use physicTick instead of physicsTick. physicTick still works but is deprecated in mineflayer.

PR #325, opened 2023-06-26
ability to go through doors, with considerations

LetsChill:
The previous person who made this change has never came back to it. This pull request brings back that feature with a lot more. continuation of #84 

There is still work to do and currently I got no idea how to implement it.

## dilemma
When the door is open, sometimes the bot get stuck and unable to move, this is because an open door is treated like air block, there need to be a function specifically for going through doors.

That function will specifically tell the bot to go the opposite direction of the door's occupied space, so it doesn't collide it.

the free space that an open door has depend on the direction and the facing of the door. but generally the length of that space is 0.0212 block only, which is a small number to get through, this needs accurate calculations.

The pathfinder sometimes has the ability to go through it with no problems, thanks to the fullStop function that centers the bot exactly in the middle of the block.

## empty spaces can start like this:

[0.0488 -> 0.0700] or [0.0300 -> 0.0512]
Those numbers are always like that, which one is present depends on the facing and the direction of the door.


IceTank:
>  this is because an open door is treated like air block

I don't think this is suppose to be the case. Doors should be treated as solid blocks at the moment.

> The pathfinder sometimes has the ability to go through it with no problems, thanks to the fullStop function that centers the bot exactly in the middle of the block.

Full stop should only apply when the bot stops and should not be used when trying to path somewhere. A better Approach is to place a node inside of the door the bot can follow.

How I would do it is to change the coordinate of the node the bot should walk towards. This function https://github.com/PrismarineJS/mineflayer-pathfinder/blob/367205e7e31e39cba7e3df09d45fcde720667846/index.js#L168 takes the calculated path from A* and adapts it so the bot can walk along it. In this function I would make a new check for a door block. And then determent where the node has to be offset so that the bot can walk through the door. So for instance:

If you have a door at position `5, 1, 5` then the center of the block is at `5.5, 1, 5.5` (Block center). (The A* result returns only full numbers and post process adds the `+ 0.5` to each node).
Now we need to figure out in what direction the door is opened. For this we can make a map of door states and door direction. So for instance door has its main hitbox towards -x. Then we add `0.1` or `0.2` to the node x value. Making it something like `5.6, 1, 5.5` (New node position). This should make the bot walk through the door without issues.

image

For figuring out in what direction the door is opened you might be able to use the block properties. https://github.com/PrismarineJS/prismarine-block/blob/master/doc/API.md#blockgetproperties this should give you the direction the door is in and if it is opened or closed. This should make it easier to get the door state across different versions instead of having to map the block stateId.

LetsChill:
Okay, I understand now, I had doubts in my ways and didn't really fully understand how path-finding completely works, thank you for this great clarification and advices

thebigsleepjoe:
Is there any progress on this PR being merged into master? I think it would be pretty useful.

rom1504:
hi @LetsChill want to have another look here?

LetsChill:
I don’t know mate I’ve been busy with major stuff lately

PR #330, opened 2023-08-20
Fix block update resets for optimized paths

m000z0rz:
On a block update, the current path gets reset if the block updated is near one of the path nodes.

On optimized paths, the distance between path nodes may be significant. Checking only the path nodes themselves may mean that a block update along the line between two path nodes gets ignored when it could impact the path.

Consider a bot that farms pumpkins - after breaking many blocks and picking them up, it could try to navigate to the next part of the farm to harvest - across flat open land, this could easily be a straight line that optimizes nicely. But, a pumpkin could suddenly grow and block the bot. Currently, if the pumpkin didn't grow right at the beginning or end of that path segment, it would be ignored, and the bot may get stuck on the newly grown pumpkin.

Improve isPositionNearPath to consider the space between path nodes by comparing the given position to the closest point along the path segment instead of just the current node.

Two optimizations keep this fast:

1) If the segment is short enough (x / y / z deltas are 2 or under) then we fall back to the previous logic where we just compare the block update position to the current node - this looks like an unoptimized path, so the old fast logic suffices.

2) If the segment looks optimized, we first check pos against the AABB enclosing the path segment. If it isn't near the AABB, then it isn't near the segment.

### DIFF
--- a/index.js
+++ b/index.js
@@ -235,13 +235,68 @@ function inject (bot) {
   }
 
   function isPositionNearPath (pos, path) {
+    let prevNode = null
     for (const node of path) {
-      const dx = Math.abs(node.x - pos.x - 0.5)
-      const dy = Math.abs(node.y - pos.y - 0.5)
-      const dz = Math.abs(node.z - pos.z - 0.5)
+      let comparisonPoint = null
+      if (
+        prevNode === null ||
+        (
+          Math.abs(prevNode.x - node.x) <= 2 &&
+          Math.abs(prevNode.y - node.y) <= 2 &&
+          Math.abs(prevNode.z - node.z) <= 2
+        )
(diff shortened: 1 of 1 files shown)


IceTank:
Ok looks good. I presume this also fixes a potential issue when the bot plans on making a sprint jump over a larger hole. Have you actually try this in combination with optimized paths? I though optimizing paths was somewhat broken.

m000z0rz:
I'm not familiar with the sprint jump over a larger hole issue. I have been using this in combination with optimized paths. I've been using them mostly because they were the default, but I do really like the movement and plan to keep using it. I've seenother issues a couple other issues, but in a mostly-open open area with underground storage it's actually working quite well for me!

PR #338, opened 2024-05-22
fix pathfinder not sprint jumping

qwqtoday:


qwqtoday:
umm what, this doesn't seem like to be my fault

IceTank:
What does this change about the behavior? Does it just sprint jump over block gaps or also on flat terrain?

qwqtoday:
> What does this change about the behavior? Does it just sprint jump over block gaps or also on flat terrain?

if you look at the original code you will see this

qwqtoday:
Oops i accidentally closed it

qwqtoday:
> What does this change about the behavior? Does it just sprint jump over block gaps or also on flat terrain?

If you look at the original code you will see this
```js
if (bot.entity.isInWater) {
      bot.setControlState('jump', true)
      bot.setControlState('sprint', false)
    } else if (stateMovements.allowSprinting && physics.canStraightLine(path, true)) {
      bot.setControlState('jump', false)
      bot.setControlState('sprint', true)
    } else if (stateMovements.allowSprinting && physics.canSprintJump(path)) {
      bot.setControlState('jump', true)
      bot.setControlState('sprint', true)
    } else if (physics.canStraightLine(path)) {
      bot.setControlState('jump', false)
      bot.setControlState('sprint', false)
    } else if (physics.canWalkJump(path)) {
      bot.setControlState('jump', true)
      bot.setControlState('sprint', false)
    } else {
      bot.setControlState('forward', false)
      bot.setControlState('sprint', false)
    }
```
The code first check is the bot in water, and it does make the bot start jumping but not sprinting, but if it’s something else like sprinting is allowed and the bot can go straight line to that direction, it makes the bot start sprinting without jumping, after that it checks can the bot sprint jumping and is sprinting allowed, but wait, if the bot can go sprint jump to that location, it also means that it can go straight line into that location. But the code checks the bot goes straight line first, so it will skip everything and never starts sprint jumping.
My change solved it by first let the bot check can it sprint jump then it let the bot sprint jump then check can it go straight line

qwqtoday:
這是一個if else if先後次序問題，原本是先檢查不可以直線走過去，然後就檢查是不是可以用跑跳過去，能跑跳過去即代表可以直線走過去，先檢查服務可否直線走過去意味着會直接跳過檢查跑跳過去

PR #340, opened 2024-09-06
fix: (spelling) scafold -> scaffold

liquidsquid1:
There has been a spelling mistake for quite some time in the code project, this pull request aims to resolve the spelling mistake so people can use proper English in their code without confusion.


liquidsquid1:
Pull request will be in draft as I need to fix compiling errors. Bear with me.

IceTank:
This is unfortunately a breaking API change and should wait until a new major release is made. Changing it now will break every library or program that currently uses pathfinder and modifies the allowed list of blocks to use.

PR #356, opened 2025-12-29
Pillar jumping bug fix

### LINKED ISSUE #296: Bot randomly fails to place blocks when scaffolding
scafolding (for search)

When a bot is bridging to a goal and has to place blocks (e.g. when traveling over void) it sometimes randomly fails to place a block but continues to move, which often causes it to fall down (and place blocks mid-air, #181).

I was able to reduce the occurences of this issue by removing
https://github.com/PrismarineJS/mineflayer-pathfinder/blob/56a02b9af715029f884cd392aa7f0836e0284c03/index.js#L500
It seems like nothing is really affected by this change except the bot is bridging slower, which is imo a fair trade off for it to not fall.
Is there another recommended solution for this issue?
Ideal solution would be to fully pause (or find another way) until the block was placed successfully, but it seems like that's not working currently.

BF5258:
\- `bot.placeBlock(refBlock, new Vec3(placingBlock.dx, placingBlock.dy, placingBlock.dz))`
\+ `bot._placeBlockWithOptions(refBlock, new Vec3(placingBlock.dx, placingBlock.dy, placingBlock.dz), { swingArm: 'right', forceLook: true })`
Place the block immediately instead of waiting for our head to turn.

\- `canPlace = placingBlock.y + 1 < bot.entity.position.y`
\+ `canPlace = placingBlock.y + 2.1 < bot.entity.position.y`
Wait until we are above the block before trying to place it. (Minecraft measures blocks from the lower north-west corner. 1 = top of original block. 2 = top of new block)

Only partially fixes #296.

### DIFF
--- a/index.js
+++ b/index.js
@@ -541,7 +541,7 @@ function inject (bot) {
       let canPlace = true
       if (placingBlock.jump) {
         bot.setControlState('jump', true)
-        canPlace = placingBlock.y + 1 < bot.entity.position.y
+        canPlace = placingBlock.y + 2.1 < bot.entity.position.y
       }
       if (canPlace) {
         if (!lockEquipItem.tryAcquire()) return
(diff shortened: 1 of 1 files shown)


vRXDk5U:
Thank you. Why has this not been merged or reviewed yet? Is there a problem?

PR #376, opened 2026-09-07
createHuman: cap the wait for the head to settle

u9g:
`settled()` resolves when the head has been quiet for `ms`, and it only ever checks that from a `physicsTick` listener. Nothing else ends the wait, so when the tick stream stops mid-gesture the promise is never answered:

- `lookAt()` awaits `settled()` directly.
- `finishWalk()` takes `settled(400).then(...).then(w.resolve, w.resolve)` on a `faceAt` walk, after it has already cleared the walk's timeout and dropped the walk, so nothing can rescue it either.

A bot that is kicked, walks out of loaded chunks, or has `physicsEnabled` turned off while a gesture is in flight leaves the caller awaiting forever. On a harness driving a controller loop that shows up as the loop simply stopping, with no error anywhere.

`settled` now also ends on wall clock, `ms + 2000` by default. The head's gestures are minimum-jerk over a few hundred milliseconds, so the cap only fires when the ticks that would have moved it are gone.

Test in `test/internalTest.js` starts a `lookAt`, drops `physicsEnabled`, and awaits it; on master it fails with mocha's timeout.

### DIFF
--- a/lib/human.js
+++ b/lib/human.js
@@ -410,17 +410,22 @@ function createHuman (bot, opts = {}) {
     w.reject(e)
   }
 
-  // Resolves once the head has stopped moving for `ms`.
-  function settled (ms) {
+  // Resolves once the head has stopped moving for `ms`. The head only moves on a physics tick, so the
+  // wait is also capped on wall clock: a kick, an unloaded chunk or physicsEnabled = false ends the
+  // tick stream, and a caller awaiting the head would otherwise never be answered at all.
+  function settled (ms, cap = ms + 2000) {
     return new Promise(resolve => {
       let quietSince = Date.now()
+      const done = () => {
+        clearTimeout(timer)
+        bot.off('physicsTick', check)
+        resolve()
+      }
       const check = () => {
         if (lastRotationDelta > 0.02 * DEG) quietSince = Date.now()
-        if (Date.now() - quietSince >= ms) {
-          bot.off('physicsTick', check)
-          resolve()
-        }
+        if (Date.now() - quietSince >= ms) done()
       }
+      const timer = setTimeout(done, cap)
       bot.on('physicsTick', check)
     })
   }
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -1308,6 +1308,20 @@ describe('human walker', function () {
     assert.strictEqual(bot.entity.pitch, 0.25, 'pitch was pulled back after lookAt')
   })
 
+  it('lookAt settles when the ticks that move the head stop', async function () {
+    this.timeout(10000)
+    bot.entity.position = spawnPos.clone()
+    const pending = human.lookAt(faceAt)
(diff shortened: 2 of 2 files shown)


lib/human.js
```diff
@@ -410,17 +410,22 @@ function createHuman (bot, opts = {}) {
...
+      }
       const check = () => {
         if (lastRotationDelta > 0.02 * DEG) quietSince = Date.now()
-        if (Date.now() - quietSince >= ms) {
-          bot.off('physicsTick', check)
-          resolve()
-        }
+        if (Date.now() - quietSince >= ms) done()
       }
+      const timer = setTimeout(done, cap)
```
rom1504, line 428: **Astra agent review — AI-generated, not manually written by the maintainer.**

The wall-clock cap currently reports the same success as a settled head. I reproduced `lookAt(point, { settleMs: 20 })` with no physics ticks: it resolved after about 2 seconds with zero `bot.look` calls and an unchanged yaw. This conflicts with the documented promise that `lookAt` (and a `faceAt` walk) resolves after the look settles, so an awaiting caller can proceed without ever facing its target. Could cap expiry be surfaced distinctly and handled by both callers, or, if best-effort completion is intentional, explicitly documented as such in the public API? Please add an assertion covering the chosen timeout outcome; the new test only checks that the promise ends.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current diff, public API documentation and both `settled()` callers, and reproduced the timeout outcome locally using the unmodified current-head controller with an event-emitter bot stub. The inline point asks for an explicit timeout contract. I did not run the server integration suite.

lib/human.js
u9g: Went with surfacing it: `settled()` now resolves `false` when the cap fires, and both `lookAt()` and the `faceAt` branch of a walk reject with `head did not settle` instead of resolving. The readme and typings document the rejection, and the tests now assert it for `lookAt` and for a `faceAt` walk (both fail with "Missing expected rejection" against the previous commit). Fixed in e5fbd00.

rom1504 (commented):

The timeout-contract finding is fixed at e5fbd00. Both no-tick regression tests pass with the actual controller and local NMP client/server fixture: lookAt and a faceAt walk reject instead of reporting successful settlement. An independent controller probe also confirms the faceAt rejection when rotation stops. Docs and the WalkOptions declaration describe the rejection.

The code looks ready from this review; please get a normal test CI run before merging. The current head's SUCCESS rollup contains only Socket Security checks, not a controller test build. I did not rerun a live vanilla server.

Skills used: `prismarine-geometry-movement-review` checked the public completion contract; `prismarine-lifecycle-action-review` checked both callers and timer/listener cleanup; `prismarine-review` verified the prior finding and the actual CI coverage.

rom1504:
Resolve conflicts and merge

PR #377, opened 2026-09-08
createHuman: a walkTo for the goal in flight joins that walk instead of superseding it

u9g:
Every \`human.walkTo\` call failed the walk in flight with \`superseded\` and started a fresh plan. A caller that re-targets faster than walks complete (an item chaser, a guard loop) got no movement at all: each call killed the previous walk while it was still planning, planning restarted, and the bot never took a step. Observed live: re-calling \`walkTo\` every 600 ms toward a pile 8 blocks away left the position unchanged for 90 s.

\`walkTo\` now returns the promise of the walk in flight when the goal is within half a block of it, planning phase included; any other goal still supersedes. The walk that finishes is not handed out again: the next call for the same goal starts a new walk.

Test: reissue the goal while the search is still slicing and then every 100 ms until the walk ends; every call returns the first promise, the bot arrives, and a call after the walk finished starts a new one. Fails on master on the first reissue.

Reported in u9g/bot-harness#74.


lib/human.js
```diff
@@ -426,8 +426,24 @@ function createHuman (bot, opts = {}) {
...
   let planSeq = 0
+  // The walkTo in flight, from the call until its promise settles: planning included, so a goal
+  // reissued while the search is still slicing joins it too.
+  let inflight = null
+
+  // A walkTo for the goal already in flight (within half a block) returns that walk's promise; any
+  // other goal supersedes it. Re-targeting on a timer would otherwise fail every walk before it
+  // took a step.
+  function walkTo (goal, o = {}) {
+    if (inflight && inflight.goal.distanceTo(goal) <= 0.5) return inflight.promise
```
rom1504, line 437: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please include the effective walk options in the equivalence check before sharing this promise. Requests for the same point can have different `radius`, `faceAt`, or `timeout` contracts. In a local reproduction, `walkTo(goal, { radius: 5 })` followed immediately by `walkTo(goal, { radius: 0.1 })` returned the same promise and both resolved while the bot was still 2 blocks from the goal. Likewise, a newly supplied `faceAt` is silently discarded. Keep joining equivalent repeated requests, but supersede or explicitly update the operation when its options differ, and cover differing options in the regression test.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current diff, option handling, documentation and tests, and reproduced option coalescing locally using the unmodified current-head controller with a minimal bot/pathfinder stub: a radius-0.1 request resolved 2 blocks away by joining a radius-5 request. I did not run the server integration suite.

lib/human.js
rom1504, line 437: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please invalidate the join record synchronously when a walk is cancelled. If an active walk is stopped and the caller immediately runs `human.walkTo(goal)` again in the same turn, this returns the stopped walk's promise: `inflight` is only cleared by a later promise callback. With the same goal and options, the actual head controller returned the original promise and the restart rejected with `stopped`; the base controller created a new walk that resolved. Reissuing during active planning still joins correctly, and waiting for the cancellation's promise cleanup before restarting also works. Keep those active joins, but make `human.stop(); human.walkTo(goal)` start fresh work, and cover this cancellation boundary.

Skills used: prismarine-lifecycle-action-review helped trace synchronous cancellation against deferred promise cleanup; prismarine-geometry-movement-review helped check the walk's public completion/cancellation contract; prismarine-review helped distinguish this from the existing differing-options feedback.

rom1504 (commented):

A synchronous stop followed by a same-goal restart can reuse the cancelled walk's promise. I compared the unmodified head and base controllers with an EventEmitter/planner adapter and checked active-join and post-cleanup restart controls. No live server or full suite was run.

Skills used: prismarine-lifecycle-action-review helped test cancellation and promise cleanup ordering; prismarine-geometry-movement-review helped check the caller's walk contract; prismarine-review helped compare current discussions and avoid duplicate feedback.

lib/human.js
u9g: done b87e689c813b9c92446cc81ecc6d5f27648c7d1b

u9g: done b87e689c813b9c92446cc81ecc6d5f27648c7d1b

PR #379, opened 2026-09-09
createHuman: hold the controller through a setback storm

u9g:
Invariants of the setback guard in `createHuman`:

- `trip` forced moves (mineflayer `forcedMove`) within `window` ms end the walk in flight with `setback`, release the controls and the head, and hold the controller for `hold` ms. Defaults: 3 in 2.5 s, 8 s hold.
- The hold ends only after `quiet` ms (3 s) without a forced move and with the bot on the ground. It never ends earlier, so the controller cannot stop and resume on each correction.
- `walkTo` and `lookAt` reject with `setback` while the hold lasts, including a walk whose planning finishes inside the hold. `human.held` reports the state.
- `setback: false` disables the guard; the head still follows a forced rotation.

Tests (`human walker`): a storm during a walk rejects it with `setback`, leaves every control released, holds later `walkTo`/`lookAt` calls, and releases once the server has been quiet; a storm during planning rejects the walk without driving the controls.

Readme and `index.d.ts` carry the option, the rejection and `human.held`.


lib/human.js
```diff
@@ -460,6 +469,7 @@ function createHuman (bot, opts = {}) {
   }
 
   async function lookAt (point, o = {}) {
+    if (heldUntil > 0) throw new Error('setback')
```
rom1504, line 472: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please recheck the hold after the reaction-delay `await`, before assigning the head target. A `lookAt()` that starts just before three `forcedMove` events passes this check, then resumes inside the new hold and reclaims the head that the guard just released. I reproduced this with the current controller: the pending call made 8 `bot.look` calls and resolved while `human.held` was still true. This defeats the new promise that the controller remains held through the storm. Add a regression that starts `lookAt`, trips the guard during its reaction delay, and verifies a `setback` rejection with no subsequent rotation.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current guard, its callers and tests, and reproduced the reaction-delay race locally using the unmodified current-head controller with an event-emitter bot stub. A look begun before the storm rotated and resolved during the hold. I did not run the server integration suite.

lib/human.js
u9g: Fixed in 5a08e4c98d5e841013b2ad7a1da799a992fb1e9e. `lookAt` now rejects with `setback` if the guard tripped during the reaction delay (checked before the head target is set) or while the head was settling. Added the regression test you described: a `lookAt` started just before three `forcedMove`s rejects with `setback` and never calls `bot.look`. It fails without the fix and passes with it.

rom1504 (commented):

Looks ready to merge from this review at 0e2bf3d. The trip-generation check after the reaction delay closes the previous race, and the second check prevents an in-progress look from reporting success after a trip. All three focused setback tests pass using the actual controller and local NMP client/server fixture: active walk interruption, planning across a hold and lookAt interrupted during its reaction delay. The current build is green. No live vanilla-server run is claimed.

Skills used: `prismarine-lifecycle-action-review` traced ownership across the await and guard trip; `prismarine-geometry-movement-review` checked the controller's observable completion/rotation behavior; `prismarine-review` verified the existing finding is addressed without duplicating it.

PR #382, opened 2026-09-12
createHuman: bridge over gaps the way a player does

u9g:
## Why the human walker cannot cross a gap today

`walkTo` plans with a Movements instance that does not place blocks, so a gap in the ground is the end of the route. Doing it by hand does not work either — this is the one placement a bot cannot make by aiming alone.

A bridge extends from the block **under the bot's feet**, and the face to click is one of that block's sides. From on top of the block no side face is in view: the ray from the eye to any point on a side face passes through the block's own top face first. The vanilla client can only ever report the face its crosshair hit (`Minecraft.pick` produces the `BlockHitResult`, and `useItemOn` is handed that object — the face and the cursor are derived from the look, never chosen alongside it), so a `use_item_on` claiming the side face from up there is a packet no client can produce.

Servers that validate the hit drop it silently. On a production BedWars server (26.1) that is every bridging placement:

```js
const edge = bot.blockAt(new Vec3(-3, 58, -67))              // the block the bot is standing on
await bot.placeBlock(edge, new Vec3(0, 0, 1))                // Server refused ... the block is still air
await bot.placeBlock(bot.blockAt(edge.offset(0, 0, -1)), new Vec3(0, 1, 0))   // ok
await bot.placeBlock(edge, new Vec3(0, 0, 1))                // Server refused ... the block is still air
```

Ten side-face attempts across three edges on two maps: ten refusals. The top face, from the same spot and the same second, places.

## What players do, and what this adds

They sneak. Sneaking lets the hitbox hang over the lip without walking off, which puts the eye horizontally *outside* the block's column, and from there the side face is the first thing the crosshair meets.

`human.bridgeTo(goal, options)` walks toward a goal one block at a time; wherever there is nothing to walk on it sneaks to the lip, clicks the face that is now in view, and steps onto what it placed. Ground the bot can step or drop down onto is walked over, so only a real gap is built across, and the bridge stays at the bot's own feet level. `human.placeAhead(direction)` is the single step on its own.

```js
const human = createHuman(bot, { seed: 42 })
await human.walkTo(islandEdge)
await human.bridgeTo(new Vec3(-2.5, 57, -62.5))   // 20 blocks of void
```

Same server and map that refused all ten: **25 blocks of bridge in one run**, the bot's first time off its island.

## Tests

Two cases in `internalTest.js`, against a server that accepts a placement only when the face the packet reports is the one the bot's own crosshair ray reaches first — all a vanilla client can send, and what a validating server checks:

- `claiming a side face from on top of the block is a hit the crosshair never makes` — head aimed at the middle of the south face and already on the wire, so only the face is under test: the packet claims south, the ray is on the top face, the server drops it.
- `bridging sneaks to the lip, so the face is the one the crosshair hits` — `bridgeTo` crosses the ledge with **no refusals at all**, every block at floor level, and the bot ends up past it.

`npm run lint` is clean. `mocha` goes from 56 passing to 58, with the one pre-existing `Physics test / simulateUntil` failure unchanged on master.


lib/human.js
```diff
@@ -467,6 +467,125 @@ function createHuman (bot, opts = {}) {
...
+    }
+    bot.setControlState('forward', false)
+    releaseHead()
+    return arrived
+  }
+
+  // Bridge toward a goal, one block at a time, along whichever axis is furthest from it. Ground that
+  // is already there is walked over; anything else is built. The bot keeps its own feet level.
+  async function bridgeTo (goal, o = {}) {
+    if (walk) failWalk(walk, new Error('superseded'))
```
rom1504, line 562: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please register bridging/standalone placement in the controller's operation lifecycle, with cancellation shared by `stop()`, `walkTo()` and subsequent bridges. This only cancels an already-created `walk`; the bridge itself is never recorded, and `stop()` neither sets `human.active = false` nor releases controls when `walk` is null. I reproduced starting a bridge and calling `stop()` during `stepOnto`: `forward` remained true. A walk still planning at entry can also become active afterward, and a newer walk/bridge can race the old bridge for the same head and controls. Check operation ownership after the awaits (including equip/aiming), and guard cleanup so an old operation cannot release its successor's controls. Please cover stop mid-step and supersession in both directions.

lib/human.js
```diff
@@ -467,6 +467,125 @@ function createHuman (bot, opts = {}) {
...
+  // on a side face passes through the block's own top face first, so a vanilla client's crosshair
+  // reports the top face and never the side. A placement that claims the side anyway is a packet no
+  // client can produce, and a server that checks the hit drops it without a word.
+  //
+  // Players get the face in view by sneaking: the hitbox may hang over the lip, which puts the eye
+  // horizontally outside the block's column, and from there the side face is the first thing the ray
+  // meets. So that is the move — sneak to the lip, click, step on.
+
+  function tickPassed () {
+    return new Promise(resolve => bot.once('physicsTick', resolve))
```
rom1504, line 483: **Astra agent review — AI-generated, not manually written by the maintainer.**

This wait has no wall-clock deadline, so the `Date.now() < until` checks in `toLip` and `stepOnto` cannot enforce their limits if physics ticks stop while awaiting it. In a local reproduction with `stepMs: 50`, stopping ticks after `forward` became true left the bridge pending and forward held past 150 ms; without another tick there is no path to the deadline check or cleanup. Give these waits a timer and cancellation path that remove their listener, and use `finally` cleanup for the owned forward/sneak/head state. Add a test that disables physics in the middle of a step and requires the operation to end within its configured bound.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the bridge implementation, controller stop/planning paths, new tests and the author's live-server/CI notes. Using the unmodified current-head controller with a minimal bot stub, I reproduced stop leaving forward held and a physics-tick wait exceeding `stepMs`. The two inline points concern operation ownership and enforceable deadlines; I did not run the server integration suite.

lib/human.js
```diff
@@ -467,6 +467,125 @@ function createHuman (bot, opts = {}) {
...
+      // and only there does sneaking hold the bot at a lip with the face in view.
+      const below = ahead.offset(0, -1, 0)
+      const footing = [0, 1, 2, 3].some(d => solid(below.offset(0, -d, 0)))
+      if (!footing) await placeAhead(dir, o)
+      if (!await stepOnto(ahead, o.stepMs ?? 1600)) {
+        // Standing over the lip leaves the feet in the block behind; back onto its middle and retry.
+        if (!await stepOnto(feet, o.stepMs ?? 1600)) throw new Error(`stuck bridging at ${bot.entity.position}`)
+      }
+    }
+    throw new Error(`${max} blocks was not enough to reach ${goal}`)
```
rom1504, line 586: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please check arrival after the final permitted step before throwing. With `blocks: 1`, the loop checks distance only before moving, so even a successful one-step trip always reaches this error. I ran the unmodified controller with released prismarine-physics on a flat floor: starting at `(0.5, 1, 0.5)`, `bridgeTo((1.5, 1, 0.5), { blocks: 1, radius: 0.6 })` ended 0.168 blocks from the goal but rejected with `1 blocks was not enough`. The same happens whenever arrival takes exactly `blocks` iterations. Check the goal after stepping (or before this throw), and cover success on the final allowed step alongside budget exhaustion.

Skills used: prismarine-behavior-test-review helped test the public completion contract at its boundary; prismarine-review helped separate this from the existing lifecycle feedback.

rom1504 (commented):

I found one additional completion-boundary issue after comparing the current diff and all prior feedback. The current-head controller with released prismarine-physics 1.11.1 reaches a one-step goal but rejects when blocks is 1. This was a local simulation on a flat full-cube floor; I did not run the server suite.

Skills used: prismarine-behavior-test-review helped check the observable completion result at the step limit; prismarine-review helped distinguish it from the existing cancellation and tick-deadline findings.

lib/human.js
u9g: Fixed in d8ce813. A walk, a bridge and a lone `placeAhead` are now one kind of operation, and only one owns the controls and head at a time ([`begin`/`cancel`/`end`/`run`](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/lib/human.js#L392-L439)). Starting another one, or `stop()`, cancels the current one: it releases the controls (sneak included) and head right away and rejects the pending waits. Every control or head write goes through an ownership check, so an old bridge that wakes after equip, aiming or a tick throws instead of touching its successor's controls. The `finally` cleanup does nothing once ownership has passed on. `walkTo` joins the same lifecycle, so a walk that is still planning when a bridge starts rejects with `superseded` and never becomes active. I left `stop()` alone on `human.active`: the readme defines `active = false` as suspending the controller, and `stop()` is meant to leave it usable for the next order.

New tests: [stop mid-step and at the lip](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/test/internalTest.js#L1529-L1550) and [supersession in both directions](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/test/internalTest.js#L1552-L1577).

lib/human.js
u9g: Fixed in d8ce813. All waits in the bridge now go through [`untilTick`](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/lib/human.js#L441-L462). It has a timer deadline, removes its physicsTick listener on every exit (tick, timeout or cancel) and rejects on cancellation. `toLip` and `stepOnto` bound their loops with it, and head aiming fails with `the head did not settle` after its settle time plus 2 s. Owned forward/sneak/head state is released by the operation's `finally` (guarded by ownership). [New test](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/test/internalTest.js#L1579-L1593): it sets `physicsEnabled = false` mid-step with `stepMs: 300`. The bridge now rejects within the bound and forward is released. Before this change it stayed pending.

lib/human.js
u9g: Fixed in d8ce813. The [loop now checks arrival before the budget](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/lib/human.js#L628-L633), so the goal is checked after the last permitted step too. [New test](https://github.com/u9g/mineflayer-pathfinder/blob/d8ce813c0b209264a32296250917516fd897f6ec/test/internalTest.js#L1595-L1602): `blocks: 1` with a one-step goal resolves, and `blocks: 1` with a goal further away still rejects with `1 blocks was not enough`.

lib/human.js
```diff
@@ -398,9 +484,10 @@ function createHuman (bot, opts = {}) {
     if (w.faceAt) {
       targetYaw = yawTo(bot.entity.position, w.faceAt)
       targetPitch = pitchTo(bot.entity.position, w.faceAt)
-      settled(400).then(releaseHead, releaseHead).then(w.resolve, w.resolve)
+      // Superseded while turning to face: the walk itself arrived, so it still resolves.
+      settled(w.op, 400).then(() => end(w.op), () => {}).then(w.resolve)
```
rom1504, line 488: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please propagate the false result from the new capped `settled()` instead of resolving this walk. With this head's unmodified controller, I start a walk already at its goal with `faceAt` pointing elsewhere, emit the arrival tick, then stop ticks. After the 2.4 s settlement cap, the walk resolves even though there were zero further rotations and the head remains short of the target. The same controlled-bot probe against the updated #376 correctly rejects with `head did not settle`. This is distinct from the fixed final bridge-step budget issue: it is the existing faceAt caller consuming the new timeout result.

Handle `ok === false` as rejection while retaining ownership-guarded cleanup, and cover stopped ticks during a faceAt walk. This was a production-controller probe with a controlled bot, not a live-server run.

Skills used: `prismarine-lifecycle-action-review` traced the changed promise result into its caller; `prismarine-geometry-movement-review` checked the documented completion contract; `prismarine-review` compared the existing threads and #376 repair.

rom1504 (commented):

The three earlier bridge findings are addressed at 7f4431d: shared operation ownership, cancellable wall-clock waits and checking arrival after the last allowed step. All eight bridging tests pass locally with the current controller and NMP client/server fixture.

One new issue remains in the refactored faceAt-walk completion: the new settlement cap returns false, but finishWalk ignores it and reports success without completing the turn. See the inline reproduction. CI also still fails in Physics/simulateUntil; the new bridge tests pass there. I have not rerun a vanilla server or independently established that CI failure at the current base.

Skills used: `prismarine-lifecycle-action-review` checked ownership and the new settlement result through every caller; `prismarine-geometry-movement-review` checked final-step arrival and observable turn completion; `prismarine-review` separated the fixed bridge threads from this newly introduced timeout result.

